PR #64, opened 2015-02-22
strtoull(3) usage walk-around for the libc of NetBSD

krytarowski:
strtoull(3) is a part of POSIX 2008 and included with: _POSIX_C_SOURCE 200809L

The current NetBSD LibC (as of pre 8.x release) enables strtoull(3) with:
- _ISOC99_SOURCE
- __STDC_VERSION__ >= 199901L
- _NETBSD_SOURCE
- __cplusplus >= 201103L

Go for -std=c99.

Please merge this with radare2.


krytarowski:
NetBSD issues reported upstream
http://gnats.netbsd.org/cgi-bin/query-pr-single.pl?number=49685

krytarowski:
The NetBSD team reports that this is fault of missing C99 defined here. So my issue is valid not just NetBSD. I will update the revision log.

radare:
I think js0n.c uses some c99 thing too, so its ok to enforce that i guess

radare:
Thanks!

PR #66, opened 2015-03-29
Sort array to fix #2188 from radare

alvarofe:


radare:
An add_sorted method would perform much better here. I have to review this change more carefully because i have some refactoring plans for sdb. Thanks!

Issue #59, opened 2014-11-12
Implement sdb_array_num_sort()


radare:
And maybe sdb_array_sort() too

radare:
Done in https://github.com/radare/sdb/pull/71

PR #67, opened 2015-03-29
Added array sorting functions #59

karel-:
didn't know alvarofe was already implementing this. but since this is a lot faster without sdb_array_get and sdb_array_set, maybe it's of use.

https://gist.github.com/karel-/74e671e91b09fb8e3f03#file-gistfile1-txt

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -18,33 +18,72 @@ static const char *Aindexof(const char *str, int idx) {
 	return NULL;
 }
 
-static const char *Aconst_index(const char *str, int idx) {
-	int len = 0;
-	const char *n, *p = str;
-	for (len=0; ; len++) {
-		if (len == idx)
-			return p;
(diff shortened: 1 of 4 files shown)


radare:
Cool! those new APIs seems interesting to have, can you please add the array_add_sorted and array_add_num_sorted() functions too?

Also, see the comments in the code style.

radare:
Can you split that pullreq into two different ones?

First with the addition of sdb_get_len() and optimizing the places where sdb_get()+strlen() is used

Second, with the array sorting apis

radare:
See `sdb_uncat` for example

radare:
I think sdb_uncat is the only function that benefits from sdb_get_len(). but there are other places which are not yet using sdb_const_get_len() like in ` sdb_json_set` with the `js` variable

karel-:
Splitting the pull request.

PR #68, opened 2015-03-30
Sort Array to fix #2188 from radare

alvarofe:
I close the PR yesterday since I didn't test properly and also I've added a sort method to num in case if necessary in the future.


radare:
I think @karel- 's implementation should perform better. I'm waiting for an updated pullreq, but i think its better to serialize the string into a native array with sdb_fmt apis, and then sort it by just swapping pointers

alvarofe:
The swapping pointers sounds better. I spoke with @karel- yesterday about the issue 2188 and I hope he continue working on the implementation

Issue #50, opened 2014-07-21
Show git commit number in version string


radare:

radare:
i did a pr from your branch https://github.com/radare/sdb/pull/172

PR #69, opened 2015-03-30
show git info with -v #50

karel-:
$ sdb -v
    sdb 0.9.4 603 git.0.9.4-11-gfbe5270
    commit: fbe527012f521131b6ee722c325b540e0ce7f4d0 build: 2015-03-30

didn't define values in sdb-version.h to avoid having to recompile everything in src after commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ test/verypush
 src/libsdb.so*
 src/sdb
 src/sdb-version.h
+src/sdb-version.c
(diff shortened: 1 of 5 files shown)


radare:
This will not work when compiling the tarball generated by `make dist`

karel-:
Maybe this need a different approach. Works when compiling the tarball generated by `make dist`. When compiling r2, sdb will get r2 version.

PR #70, opened 2015-03-30
Add sdb_get_len()

karel-:
First part of the split pull request #67.

Added `sdb_get_len` and changed `sdb_uncat` to use it instead of `sdb_get` and `strlen`.
Changed `sdb_json_set` to use `sdb_const_get_len`.
Created new `sdb_alen_ignore_empty` for using instead of `sdb_get` and `sdb_array_compact` in `sdb_arry_length` (previous variant had unnecessary malloc and memmove, as the compact array itself was discarded).


radare:
Thanks!

PR #71, opened 2015-03-30
Added sdb_array_sort() and sdb_array_sort_num()

karel-:
Second part of the split pull request #67. Completes #59.
Implemented `sdb_array_sort` and `sdb_array_sort_num`. Will add `sdb_array_add_sorted` later. @alvarofe can maybe already use these.

### DIFF
--- a/src/array.c
+++ b/src/array.c
@@ -18,34 +18,34 @@ static const char *Aindexof(const char *str, int idx) {
 	return NULL;
 }
 
-static const char *Aconst_index(const char *str, int idx) {
-	int len = 0;
-	const char *n, *p = str;
-	for (len=0; ; len++) {
-		if (len == idx)
-			return p;
-		n = strchr (p, SDB_RS);
-		if (n) p = n+1;
-		else break;
+static int astrcmp (const char *a, const char *b) {
+	register char va = *a;
+	register char vb = *b;
+	for (;;) {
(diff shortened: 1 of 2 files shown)


karel-:
@radare as per the register discussion from #67. With those swap functions using register hints didn't produce a significant improvement. Compiler chose to use rbx, but poping and pushing it kind of cancelled out it's positive effect. And compiler ignored inline hint. Why not use optimization flags?

radare:
Thanks!

radare:
Yep, there are some articles that talk about not using compiler hints with modern compilers because they are just ignored to avoid human errors. But i'm not sure if we should go to for this kind of optimizations yet. There are still several places to optimize sdb, and i'll try to work on them before 1.0, but current priority right now is to fix the remaining tests market as broken. And add more tests.

The query syntax parser is quite ugly and i would like to rewrite it after 1.0, and probalby change the syntax, but i need to think more on this and discuss, so 

Can you do the sdb_array_add_sorted() and num_sorted() functions too? in some situations is better to use sorted-adds instead of just adding them all in a shot and then sort.

Also, after 1.0 i have plans to split the API into different levels to cover backend, core, value abstractions (json, array, ranges, ..) and high level constructions (sdbtypes in C and with more complex types). This way i can stabilize and track the issues and design of each module independently and keep it open to support more helpers for working with complex data structures into sdb.

PR #73, opened 2015-04-01
Functions for adding to sorted arrays.

karel-:


radare:
Thanks! merged we need some tests on this. do you propose any syntax for sdbquery to use to this api?

karel-:
I'll look into it later, don't have a certain syntax in mind right now.

karel-:
Syntax. I suggest (add_sorted, sort)

    [!+]key=foo
    [!]key

which is kind of similar to the existing `[+]key=foo`. And perhaps

    [#+]key=num
    [#]key

as numeric counterparts. If that suits, I'll include it in sdbquery and add tests. And I have no idea whether git notifies you that I'm ranting in a closed pull request so maybe this helps - @radare

radare:
Dont reuse closed pullreqs. Make a new branch cloned from master and work there. Otherwise it can result in conflictive merges

PR #81, opened 2015-05-25
Update path.c

skuater:
fix bug in json parser


radare:
Awesome catch! Thanks

PR #92, opened 2016-01-23
Reduce CDB header size (2048 bytes -> 1024 bytes)

deffi420:
The trick is to compute the subtable length n_i, 0 <= i <= 255, using
the subtable pointers p_i and p_{i+1}. Observe that

        n_i = (p_{i+1} - p_i) / L, where L is the slot length.

So let's not waste 1024 bytes of space to encode n_i.

### DIFF
--- a/src/cdb.c
+++ b/src/cdb.c
@@ -113,16 +113,22 @@ int cdb_findnext(struct cdb *c, ut32 u, const char *key, ut32 len) {
 	if (c->fd == -1) return -1;
 	c->hslots = 0;
 	if (!c->loop) {
-		if (!cdb_read (c, buf, sizeof (buf), (u << 3) & 2047)) {
+		int bufsz = ((u+1) & 0xFF) ? sizeof (buf) : sizeof (buf) / 2;
+		if (!cdb_read (c, buf, bufsz, (u << 2) & 1023))
 			return -1;
-		}
-		ut32_unpack (buf + 4, &c->hslots);
-		if (!c->hslots) {
-			return 0;
-		}
+
+		/* hslots = (hpos_next - hpos) / 8 */
 		ut32_unpack (buf, &c->hpos);
+		if (bufsz == sizeof (buf))
+			ut32_unpack (buf+4, &pos);
+		else pos = c->size;
+
+		if (pos < c->hpos) return -1;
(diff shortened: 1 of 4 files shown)


radare:
screen shot 2016-01-23 at 23 38 01

this change makes sdb a bit slower

radare:
in percentages:

* database seems to be 15% smaller
* speed is 1% slower

radare:
For some cases it will be faster to precalculate all the `n_i`s and keep it in memory until the sdb instance is gone. But this seems to be a fair exchange between dbsize and performance

deffi420:
2016-01-24-014702_908x767_scrot
The running times are pretty much the same for me (CFLAGS="-O2"). Nonetheless, I'll try to optimize the n_i computation tomorrow.

radare:
For 100K rows , the new implementation takes 0.370s vs 0.447s to generate the database. so it means that syncing is faster, which is very important operation because its never updating the disk file. (18% faster)

and with 100K rows, the database is much smaller. 4.1MB vs 3.3MB.

We are missing a test for retrieving, i’ll add it now, so i think the benchmarks clearly show that this thing must be merged :)

Send those optimizations in a separate PR.

radare:
It’s obvious, but worthless to say that this change makes databases built with the previous code are no longer accessible. this will break for example the information in projects in r2.

as long as sdb deosnt puts any header in the database, there’s no way to specify a version number or filetype and we cant handle this. But i think we can go forward, and just warn users about this breakage, but i dont think its gonna be a huge problem

deffi420:
I'm afraid there's something wrong. The database should be 1024 bytes smaller, not 15%...

2016-01-24-022217_1359x767_scrot

For some reason the new CDB hashmap is 100% full (i.e., collision occur all the time). We need some empty slots to speed up lookups.

deffi420:
Ok, the problem is here https://github.com/radare/sdb/commit/90a6981271b611dc880ed06d4f913e04476af285#diff-707f52f9224c9ce39cb59909122d1581R150

count is the number of reserved slots in a subtable, and len is the total length of the subtable. We do not want these to be equal because then there is no empty slots and hash collisions occur 100% of the time. So we should revert the change and set: len = count << 1.

My bad...

PR #93, opened 2016-01-24
Fix previous commit

deffi420:
Explanation: https://github.com/radare/sdb/pull/92#issuecomment-174239234


radare:
Awesome :) benchmarks look a bit better after this , too

radare:
And yep. now the database is "big" again.

Issue #75, opened 2015-04-03
SDB failed to create files in the Windows from the MIngw32 shell


XVilka:
SDB failed to create files in the Windows from the MIngw32 shell

```
a.kochkov@WS-ANTON ~/radare/radare2/libr/syscall/d
$ /bin/sh gen.sh < dos-x86-16 | ../../../libr/../shlr/sdb/sdb dos-x86-16.sdb =
sdb: Cannot open 'dos-x86-16.sdb.tmp' for writing.
Cannot create database
a.kochkov@WS-ANTON ~/radare/radare2/libr/syscall/d
```

Original error from the radare2 building process

```
make[3]: Entering directory `/d/Work/radare/radare2/libr/syscall'
make -C d
make[4]: Entering directory `/d/Work/radare/radare2/libr/syscall/d'
make[5]: Entering directory `/d/Work/radare/radare2/libr/syscall/d'
rm -f dos-x86-16.sdb
/bin/sh gen.sh < dos-x86-16 | ../../../libr/../shlr/sdb/sdb dos-x86-16.sdb =
make[5]: *** [dos-x86-16.sdb] Error 116
make[5]: Leaving directory `/d/Work/radare/radare2/libr/syscall/d'
make[4]: *** [all] Error 2
make[4]: Leaving directory `/d/Work/radare/radare2/libr/syscall/d'
make[3]: *** [do] Error 2
```

radare:
works for me on wine. this is very important, so let me know how can i reproduce this from linux/osx

XVilka:
@radare so this is the regression https://github.com/radare/radare2/issues/1582

XVilka:
Was fixed in this commit https://github.com/radare/sdb/commit/874e92a4f8b89df8d33a4162fe6eb7a3e787e0d6

XVilka:
This time though the reason is another (still trying to find out).

radare:
Cc @skuater

XVilka:
@radare @skuater - any ideas how to fix this? This prevents from building radare2 on mingw32 on some windows systems.

radare:
can anyone reproduce? :)

XVilka:
I can provide the failing binary

XVilka:
Here it is http://xvilka.me/sdb.exe

radare:
Can u share build logs and compiler version/etc? Can you reproduce the same issue by building sdb from git instead of building yhe one shipped in r2?

XVilka:
[...]

PR #95, opened 2016-02-15
MinGW: use _aligned_free() instead of free()

0xABD:
As per Issue #75, for MinGW use _aligned_free() instead of free() for memory allocated with _aligned_malloc()


radare:
Woah nice catch :D thanks!

PR #99, opened 2016-06-04
Merge the implementation of sdb_get_len() with sdb_const_get_len()

dequis:
Now `sdb_get_len()` is just a wrapper for `sdb_const_get_len()` that strdups the result.

Both functions looked almost the same, just a few changes that were in one function but not the other which are now merged.


dequis:
(This is what I mentioned in https://github.com/radare/radare2/pull/5076)

radare:
Thanks! i have updated the copy in r2

PR #102, opened 2016-06-21
Rhashtable2

crowell:
replace the hashtable with one that can handle collisions.

not sure how to test this, but it works from my small amount of testing by hand...

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -57,7 +57,9 @@ SDB_API int sdb_disk_create (Sdb* s) {
 	memcpy (str, s->dir, nlen+1);
 	r_sys_mkdirp (str);
 	memcpy (str+nlen, ".tmp", 5);
-	close (s->fdump);
+	if (s->fdump != -1) {
+		close (s->fdump);
+	}
 	s->fdump = open (str, O_BINARY|O_RDWR|O_CREAT|O_TRUNC, SDB_MODE);
 	if (s->fdump == -1) {
 		eprintf ("sdb: Cannot open '%s' for writing.\n", str);
--- a/src/ht.c
+++ b/src/ht.c
@@ -1,3 +1,4 @@
+/* radare2 - BSD 3 Clause License - crowell 2016 */
 #include "ht.h"
 #include "sdb.h"
 
@@ -19,29 +20,23 @@ const int ht_primes_sizes[] = {
 // pair_free - function for freeing a keyvaluepair - if NULL just does free.
 // calcsize - functino to calculate the size of a value. if NULL, just stores 0.
 SdbHash* internal_ht_new(ut32 size, HashFunction hashfunction, ListComparator comparator, DupKey keydup, DupValue valdup, HtKvFreeFunc pair_free, CalcSize calcsize) {
-	SdbHash* ht;
 	ut32 i;
-	ht = calloc (1, sizeof (*ht));
-	if (ht) {
-		ht->size = size;
-		ht->count = 0;
-		ht->prime_idx = 0;
-		ht->load_factor = 1;
-		ht->hashfn = hashfunction;
-		ht->cmp = comparator;
-		ht->dupkey = keydup;
-		ht->dupvalue = valdup;
-		ht->table = calloc (ht->size, sizeof (SdbList*));
-		ht->calcsize = calcsize;
-		if (ht->table) {
-			for (i = 0; i < ht->size; ++i) {
-				ht->table[i] = ls_newf ((SdbListFree)pair_free);
-			}
-		} else {
-			free (ht);
-			return NULL;
-		}
+	SdbHas* ht = calloc (1, sizeof (*ht));
+	if (!ht) {
+		return NULL;
 	}
+	ht->size = size;
+	ht->count = 0;
+	ht->prime_idx = 0;
+	ht->load_factor = 1;
+	ht->hashfn = hashfunction;
+	ht->cmp = comparator;
+	ht->dupkey = keydup;
+	ht->dupvalue = valdup;
+	ht->table = calloc (ht->size, sizeof (SdbList*));
+	ht->calcsize = calcsize;
+	ht->freefn = pair_free;
+	// Because we use calloc, each listptr will be NULL until used */
 	return ht;
 }
 
@@ -84,7 +79,7 @@ SdbHash* ht_new() {
 
 void ht_free(SdbHash* ht) {
 	ut32 i;
-	for (i = 0; i < ht->size; ++i) {
+	for (i = 0; i < ht->size; i++) {
 		ls_free (ht->table[i]);
 	}
 	free (ht->table);
@@ -102,7 +97,7 @@ static void internal_ht_grow(SdbHash* ht) {
 	ht2 = internal_ht_new (sz, ht->hashfn, ht->cmp, ht->dupkey,
 			ht->dupvalue, (HtKvFreeFunc)ht->table[0]->free, ht->calcsize);
 	ht2->prime_idx = ht->prime_idx;
-	for (i = 0; i < ht->size; ++i) {
+	for (i = 0; i < ht->size; i++) {
 		ls_foreach (ht->table[i], iter, kvp) {
 			(void)ht_insert (ht2, kvp->key, kvp->value);
 		}
@@ -147,11 +142,14 @@ static bool internal_ht_insert(SdbHash* ht, bool update, const char* key, const 
 			} else {
 				kvp->value_len = 0;
 			}
+			if (!ht->table[bucket]) {
+				ht->table[bucket] = ls_newf (ht->freefn);
+			}
 			ls_prepend (ht->table[bucket], kvp);
 			ht->count++;
 			// Check if we need to grow the table.
 			if (ht->count >= ht->load_factor * ht_primes_sizes[ht->prime_idx]) {
-				if (ht->prime_idx < sizeof (ht_primes_sizes)/sizeof (ht_primes_sizes[0])) {
+				if (ht->prime_idx < sizeof (ht_primes_sizes) / sizeof (ht_primes_sizes[0])) {
 					ht->prime_idx++;
 					internal_ht_grow (ht);
 				}
@@ -182,18 +180,19 @@ bool ht_insert_kvp(SdbHash* ht, SdbKv* kvp, bool update) {
 	} else {
 		(void)ht_find (ht, kvp->key, &found);
 	}
-	if (update || !found) {
-		if (kvp) {
-			bucket = hash % ht->size;
-			ls_prepend (ht->table[bucket], kvp);
-			ht->count++;
-			// Check if we need to grow the table.
-			if (ht->count >= ht->load_factor * ht_primes_sizes[ht->prime_idx]) {
-				ht->prime_idx++;
-				internal_ht_grow (ht);
-			}
-			return true;
+	if (kvp && (update || !found)) {
+		bucket = hash % ht->size;
+		if (!ht->table[bucket]) {
+			ht->table[bucket] = ls_newf (ht->freefn);
 		}
+		ls_prepend (ht->table[bucket], kvp);
+		ht->count++;
+		// Check if we need to grow the table.
+		if (ht->count >= ht->load_factor * ht_primes_sizes[ht->prime_idx]) {
+			ht->prime_idx++;
+			internal_ht_grow (ht);
+		}
+		return true;
 	}
 	return false;
 }
--- a/src/ht.c
+++ b/src/ht.c
@@ -94,7 +94,7 @@ static void internal_ht_grow(SdbHash* ht) {
 	ut32 i;
 	ut32 sz = ht_primes_sizes[ht->prime_idx];
 	ht2 = internal_ht_new (sz, ht->hashfn, ht->cmp, ht->dupkey,
-			ht->dupvalue, (HtKvFreeFunc)ht->table[0]->free, ht->calcsize);
+			ht->dupvalue, (HtKvFreeFunc)ht->freefn, ht->calcsize);
 	ht2->prime_idx = ht->prime_idx;
 	for (i = 0; i < ht->size; i++) {
 		ls_foreach (ht->table[i], iter, kvp) {
--- a/src/ht.c
+++ b/src/ht.c
@@ -20,8 +20,7 @@ const int ht_primes_sizes[] = {
 // pair_free - function for freeing a keyvaluepair - if NULL just does free.
 // calcsize - functino to calculate the size of a value. if NULL, just stores 0.
 SdbHash* internal_ht_new(ut32 size, HashFunction hashfunction, ListComparator comparator, DupKey keydup, DupValue valdup, HtKvFreeFunc pair_free, CalcSize calcsize) {
-	ut32 i;
-	SdbHas* ht = calloc (1, sizeof (*ht));
+	SdbHash* ht = calloc (1, sizeof (*ht));
 	if (!ht) {
 		return NULL;
 	}
@@ -143,7 +142,7 @@ static bool internal_ht_insert(SdbHash* ht, bool update, const char* key, const 
 				kvp->value_len = 0;
 			}
 			if (!ht->table[bucket]) {
-				ht->table[bucket] = ls_newf (ht->freefn);
+				ht->table[bucket] = ls_newf ((SdbListFree)ht->freefn);
 			}
 			ls_prepend (ht->table[bucket], kvp);
 			ht->count++;
@@ -183,7 +182,7 @@ bool ht_insert_kvp(SdbHash* ht, SdbKv* kvp, bool update) {
 	if (kvp && (update || !found)) {
 		bucket = hash % ht->size;
 		if (!ht->table[bucket]) {
-			ht->table[bucket] = ls_newf (ht->freefn);
+			ht->table[bucket] = ls_newf ((SdbListFree)ht->freefn);
 		}
 		ls_prepend (ht->table[bucket], kvp);
 		ht->count++;
--- a/src/ht.c
+++ b/src/ht.c
@@ -1,239 +1,238 @@
-/*
- * Copyright © 2009 Intel Corporation
- * Copyright © 1988-2004 Keith Packard and Bart Massey.
- *
- * Permission is hereby granted, free of charge, to any person obtaining a
- * copy of this software and associated documentation files (the "Software"),
- * to deal in the Software without restriction, including without limitation
- * the rights to use, copy, modify, merge, publish, distribute, sublicense,
- * and/or sell copies of the Software, and to permit persons to whom the
- * Software is furnished to do so, subject to the following conditions:
- *
- * The above copyright notice and this permission notice (including the next
- * paragraph) shall be included in all copies or substantial portions of the
- * Software.
- *
- * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
- * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
- * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.  IN NO EVENT SHALL
- * THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
- * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
- * FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS
- * IN THE SOFTWARE.
- *
- * Except as contained in this notice, the names of the authors
- * or their institutions shall not be used in advertising or
- * otherwise to promote the sale, use or other dealings in this
- * Software without prior written authorization from the
- * authors.
- *
- * Authors:
- *    Eric Anholt <eric@anholt.net>
- *    Keith Packard <keithp@keithp.com>
- * Integration in r2 core api and hackit up
- *    pancake <nopcode.org>
- */
-
 #include "ht.h"
+#include "sdb.h"
 
-#define ARRAY_SIZE(array) (sizeof(array) / sizeof(array[0]))
+// Sizes of the ht.
+const int ht_primes_sizes[] = {
+3, 7, 11, 17, 23, 29, 37, 47, 59, 71, 89, 107, 131, 163, 197, 239, 293, 353, 431, 521, 631, 761, 919,
+1103, 1327, 1597, 1931, 2333, 2801, 3371, 4049, 4861, 5839, 7013, 8419, 10103, 12143, 14591,
+17519, 21023, 25229, 30293, 36353, 43627, 52361, 62851, 75431, 90523, 108631, 130363, 156437,
+187751, 225307, 270371, 324449, 389357, 467237, 560689, 672827, 807403, 968897, 1162687, 1395263,
+1674319, 2009191, 2411033, 2893249, 3471899, 4166287, 4999559, 5999471, 7199369};
 
-/*
- * From Knuth -- a good choice for hash/rehash values is p, p-2 where
- * p and p-2 are both prime.  These tables are sized to have an extra 10%
- * free to avoid exponential performance degradation as the hash table fills
- */
 
-static ut32 deleted_data;
+// Create a new hashtable and return a pointer to it.
+// size - number of buckets in the hashtable
+// hashfunction - the function that does the hashing, must not be null.
+// comparator - the function to check if values are equal, if NULL, just checks == (for storing ints).
+// keydup - function to duplicate to key (eg strdup), if NULL just does =.
+// valdup - same as keydup, but for values
+// pair_free - function for freeing a keyvaluepair - if NULL just does free.
+// calcsize - functino to calculate the size of a value. if NULL, just stores 0.
+SdbHash* internal_ht_new(ut32 size, HashFunction hashfunction, ListComparator comparator, DupKey keydup, DupValue valdup, HtKvFreeFunc pair_free, CalcSize calcsize) {
+	SdbHash* ht;
+	ut32 i;
+	ht = calloc (1, sizeof (*ht));
+	if (ht) {
+		ht->size = size;
+		ht->count = 0;
+		ht->prime_idx = 0;
+		ht->load_factor = 1;
+		ht->hashfn = hashfunction;
+		ht->cmp = comparator;
+		ht->dupkey = keydup;
+		ht->dupvalue = valdup;
+		ht->table = calloc (ht->size, sizeof (SdbList*));
+		ht->calcsize = calcsize;
+		if (ht->table) {
+			for (i = 0; i < ht->size; ++i) {
+				ht->table[i] = ls_newf ((SdbListFree)pair_free);
+			}
+		} else {
+			free (ht);
+			return NULL;
+		}
+	}
+	return ht;
+}
 
-static const struct {
-	ut32 max_entries, size, rehash;
-} hash_sizes[] = {
-	{ 2,		5,		3	  },
-	{ 4,		7,		5	  },
-	{ 8,		13,		11	  },
-	{ 16,		19,		17	  },
-	{ 32,		43,		41        },
-	{ 64,		73,		71        },
-	{ 128,		151,		149       },
-	{ 256,		283,		281       },
-	{ 512,		571,		569       },
-	{ 1024,		1153,		1151      },
-	{ 2048,		2269,		2267      },
-	{ 4096,		4519,		4517      },
-	{ 8192,		9013,		9011      },
-	{ 16384,	18043,		18041     },
-	{ 32768,	36109,		36107     },
-	{ 65536,	72091,		72089     },
-	{ 131072,	144409,		144407    },
-	{ 262144,	288361,		288359    },
-	{ 524288,	576883,		576881    },
-	{ 1048576,	1153459,	1153457   },
-	{ 2097152,	2307163,	2307161   },
-	{ 4194304,	4613893,	4613891   },
-	{ 8388608,	9227641,	9227639   },
-	{ 16777216,	18455029,	18455027  },
-	{ 33554432,	36911011,	36911009  },
-	{ 67108864,	73819861,	73819859  },
-	{ 134217728,	147639589,	147639587 },
-	{ 268435456,	295279081,	295279079 },
-	{ 536870912,	590559793,	590559791 },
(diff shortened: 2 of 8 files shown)


radare:
maybe we should remove the hashtable implementation of r2 and, as long as sdb goes into rutil now, just use the one in sdb, to avoid duppes. what do you think?

crowell:
i think that's a pretty good idea.

src/sdb.c
```diff
@@ -115,7 +115,7 @@ static void sdb_fini(Sdb* s, int donull) {
 	free (s->name);
 	free (s->path);
 	ls_free (s->ns);
-	ht_free (s->ht);
+	ht_destroy (s->ht);
```
radare: free != destroy

usually free liberates all the fields stored by the object. destroy will just free the object itself without freeing the contents

src/sdb.c
```diff
@@ -360,16 +354,18 @@ SDB_API SdbKv* sdb_kv_new (const char *k, const char *v) {
 }
 
 SDB_API void sdb_kv_free (SdbKv *kv) {
```
radare: memory leak

src/sdb.c
```diff
@@ -480,11 +474,13 @@ SDB_API int sdb_foreach (Sdb* s, SdbForeachCallback cb, void *user) {
...
-			return 0;
+	for (i = 0; i < s->ht->size; ++i) {
+		ls_foreach (s->ht->table[i], iter, kv) {
+			if (!kv->value || !*kv->value)
+				continue;
+			if (!cb (user, kv->key, kv->value))
+				return 0;
+		}
 	}
 	return 1;
```
radare: we should boolify and add more braces. aka use r2 syntax in sdb too

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+		ht->size = size;
+		ht->count = 0;
+		ht->load_factor = 1;
+		ht->hashfn = hashfunction;
+		ht->cmp = comparator;
+		ht->dupkey = keydup;
+		ht->dupvalue = valdup;
+		ht->table = calloc (ht->size, sizeof (SdbList*));
+		if (ht->table != NULL) {
+			for (i = 0; i < ht->size; ++i) {
```
radare: woah, this multiplicates the memory usage and heap fragmentation of a hahstable A LOT.

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
-				double_hash = 1;
-			hash_address = (hash_address + double_hash) % ht->size;
-		} while (hash_address != hash % ht->size);
+SdbHash* internal_ht_new(ut64 size, HashFunction hashfunction,
+		ListComparator comparator, DupKey keydup, DupValue valdup,
+		HtKvFreeFunc pair_free) {
+	SdbHash* ht;
+	ut32 i;
+	ht = calloc (1, sizeof(*ht));
+	if (ht != NULL) {
```
radare: != NULL is unnecessarsy

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
-			if (double_hash == 0)
-				double_hash = 1;
-			hash_address = (hash_address + double_hash) % ht->size;
-		} while (hash_address != hash % ht->size);
+SdbHash* internal_ht_new(ut64 size, HashFunction hashfunction,
+		ListComparator comparator, DupKey keydup, DupValue valdup,
+		HtKvFreeFunc pair_free) {
+	SdbHash* ht;
+	ut32 i;
+	ht = calloc (1, sizeof(*ht));
```
radare: missing space before (

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+	bool found = true;
+	if (update) {
+		ht_delete (ht, key);
+	} else {
+		(void)ht_find (ht, key, &found);
+	}
+	if (update || !found) {
+		kvp = calloc (1, sizeof (SdbKv *));
+		if (kvp) {
+			if (ht->dupkey) {
```
radare: use ternary operator

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
 
-void *ht_lookup(SdbHash *ht, ut32 hash) {
-	SdbHashEntry *entry;
-	if (!ht || !ht->list || !ht->list->head) return NULL;
-	entry = ht_search (ht, hash);
-	return entry? entry->data : NULL;
+// Inserts the key value pair key, value into the hashtable.
+// if update is true, allow for updates, otherwise return false if the key
+// already exists.
+static bool internal_ht_insert(SdbHash* ht, bool update, const char* key,
```
radare: do not define functions in more than one line

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+// Looks up the corresponding value from the key. Returns true if found, false
+// otherwise.
+SdbKv* ht_find_kvp(SdbHash* ht, const char* key, bool* found) {
+	ut64 hash;
+	ut64 bucket;
+	SdbListIter* iter;
+	SdbKv* kvp;
+	hash = ht->hashfn (key);
+	bucket = hash % ht->size;
+	ls_foreach (ht->table[bucket], iter, kvp) {
```
radare: a hashtable without primer numbers seems unefficient to me. did you benchmarked both implementations? memory and cpu usage is important

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
 }
 
-void ht_delete_entry(SdbHash *ht, SdbHashEntry *entry) {
-	if (!ht || !entry)
-		return;
-	if (!rehash && entry->iter) {
-		ls_delete (ht->list, entry->iter);
-		entry->iter = NULL;
+// Deletes a kvp from the hash table from the key, if the pair exists.
+void ht_delete(SdbHash* ht, const char* key) {
```
radare: make it return bool to notify the user if the key has been removed or not

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+	ut64 hash;
```
radare: you can define both vars in the same line because its same type and name is short

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+	SdbList* list = NULL;
+	SdbKv* kvp = NULL;
+	hash = ht->hashfn (key);
+	bucket = hash % ht->size;
+	list = ht->table[bucket];
+	ls_foreach (list, iter, kvp) {
+		if (ht->cmp) {
+			if (ht->cmp (key, kvp->key) == 0) {
+				ls_delete (list, iter);
+				--ht->count;
```
radare: dont pre-increment. this doesnt provides any benefit and makes reading hard

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+	SdbListIter* iter = NULL;
+	SdbList* list = NULL;
+	SdbKv* kvp = NULL;
+	hash = ht->hashfn (key);
+	bucket = hash % ht->size;
+	list = ht->table[bucket];
+	ls_foreach (list, iter, kvp) {
+		if (ht->cmp) {
+			if (ht->cmp (key, kvp->key) == 0) {
+				ls_delete (list, iter);
```
radare: maybe r2 should definitively use only Sdb structures at some point. to avoid also the RList vs SdbList

src/ht.h
```diff
@@ -1,31 +1,59 @@
...
+// RHashFunction - the function pointer to the hash function. For a string, use r_str_hash64 for dbj2. Must not be NULL.
+// comparator - the function pointer to comparison function. Returns 0 for equality eg) for a string strcmp.
+// keydup - the function pointer to a dup function. eg) for a string strdup, if NULL is provided, just assignment of the value/pointer
+// valdup - the function pointer to a dup function. eg) for a string strdup, if NULL is provided, just assignment of the value/pointer
+// pair_free - the function pointer to the function to free a SdbKv entry, if NULL is provided, just free is used.
+SdbHash* ht_new(HashFunction hashfunction, ListComparator comparator,
+	DupKey keydup, DupValue valdup, HtKvFreeFunc pair_free);
+
+// Destroy a hashtable and all of its entries.
+void ht_destroy(SdbHash* ht);
```
radare: rename to ht_free

src/sdb.c
```diff
@@ -261,18 +254,18 @@ SDB_API int sdb_add (Sdb* s, const char *key, const char *val, ut32 cas) {
...
 	char ch;
 	SdbKv *kv;
+	bool found;
 	int klen = strlen (key)+1;
 	if (!s) return 0;
-	hash = sdb_hash (key);
-	kv = (SdbKv*)ht_lookup (s->ht, hash);
-	if (kv) return (*kv->value)? 1: 0;
+	kv = (SdbKv*)ht_find_kvp (s->ht, key, &found);
+	if (found && kv) return (*kv->value)? 1: 0;
```
radare: add braces

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+	free (ht);
 }
 
-void ht_free(SdbHash *ht) {
-	if (ht) {
-		free (ht->table);
-		ls_free (ht->list);
-		free (ht);
+// Increases the size of the hashtable by 2.
+static void internal_ht_grow(SdbHash* ht) {
```
radare: totally unefficient operation

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
-			if (entry_is_deleted (entry))
-				ht->deleted_entries--;
-			entry->hash = hash;
-			entry->data = data;
-			entry->iter = rehash? iter: ls_append (ht->list, data);
-			ht->entries++;
-			return 1;
+// Looks up the corresponding value from the key. Returns true if found, false
+// otherwise.
+SdbKv* ht_find_kvp(SdbHash* ht, const char* key, bool* found) {
```
radare: oh i think i get the reason for the SdbKv here.. beacuse of cas and timeout.. the problem here is that you are mixing concepts. because those concepts are far beyond the hashtable functionality. this is. a hashtable should work independently from the contents it's storing. you are binding Sdb concepts into that hashtable implementation.

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+		if (ht->cmp) {
+			if (ht->cmp (key, kvp->key) == 0) {
+				ls_delete (list, iter);
+				--ht->count;
+				return;
+			}
+		} else {
+			if (key == kvp->key) {
+				ls_delete (list, iter);
+				--ht->count;
```
radare: dont preincrement

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+	ls_foreach (list, iter, kvp) {
+		if (ht->cmp) {
+			if (ht->cmp (key, kvp->key) == 0) {
+				ls_delete (list, iter);
+				--ht->count;
+				return;
+			}
+		} else {
+			if (key == kvp->key) {
+				ls_delete (list, iter);
```
radare: spaguetti body

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+	ut64 hash;
+	ut64 bucket;
+	SdbListIter* iter;
+	SdbKv* kvp;
+	hash = ht->hashfn (key);
+	bucket = hash % ht->size;
+	ls_foreach (ht->table[bucket], iter, kvp) {
+		if (ht->cmp) {
+			if (ht->cmp (key, kvp->key) == 0) {
+				*found = true;
```
radare: spaguetti

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+			}
+			if (ht->dupvalue) {
+				kvp->value = ht->dupvalue (value);
+			} else {
+				kvp->value = (char*)value;
+			}
+			hash = ht->hashfn (key);
+			bucket = hash % ht->size;
+			kvp->expire = 0;
+			kvp->value_len = strlen (kvp->value);
```
radare: performance loss. the value is already strlened few lines before

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+	ut64 hash;
+	ut64 bucket;
+	bool found = true;
+	if (update) {
+		ht_delete (ht, key);
+	} else {
+		(void)ht_find (ht, key, &found);
+	}
+	if (update || !found) {
+		kvp = calloc (1, sizeof (SdbKv *));
```
radare: hash table should be independent from sdb.

src/ht.c
radare: when is this grow operation happening?

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
-	ht->deleted_entries = 0;
-	for (e = old_ht.table; e != old_ht.table + old_ht.size; e++) {
-		if (entry_is_present (e))
-			ht_insert (ht, e->hash, e->data, e->iter);
-	}
-	free (old_ht.table);
-rehash = 0;
+SdbHash* ht_new(HashFunction hashfunction, ListComparator comparator,
+		DupKey keydup, DupValue valdup, HtKvFreeFunc pair_free) {
+	ut64 initial_size = 1024;
```
radare: why ut64? why 1024? use constants at the begining of the file not in th emiddle of a function

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+		ht->count = 0;
+		ht->load_factor = 1;
+		ht->hashfn = hashfunction;
+		ht->cmp = comparator;
+		ht->dupkey = keydup;
+		ht->dupvalue = valdup;
+		ht->table = calloc (ht->size, sizeof (SdbList*));
+		if (ht->table != NULL) {
+			for (i = 0; i < ht->size; ++i) {
+				ht->table[i] = ls_new ();
```
radare: you can use ls_newf instead of setting the ->free pointer later by hand

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+				if (pair_free) {
+					ht->table[i]->free = (SdbListFree)pair_free;
```
radare: null deref

src/ht.c
```diff
@@ -1,239 +1,184 @@
...
+				}
+			}
+		} else {
+			free (ht);
```
radare: loads of memory leaks

src/ht.c
crowell: no it isnt???

radare: isnt dupvalue() strlening it already?

src/sdb.c
crowell: done.

src/sdb.c
```diff
@@ -74,7 +74,7 @@ SDB_API Sdb* sdb_new (const char *path, const char *name, int lock) {
 		goto fail;
 	s->ns->free = NULL;
 	if (!s->ns) goto fail;
-	s->ht = ht_new ((SdbListFree)sdb_kv_free);
+	s->ht = ht_new (sdb_hash, strcmp, strdup, strdup, sdb_kv_free);
```
crowell: eh, it's the cost of haivng it modular, really it also needs another argument for "size" which strlen is used for...

src/ht.c
crowell: understood, but it's the cost of correctness.

src/ht.c
crowell: done

src/ht.c
crowell: will fix in next iteration;

src/ht.c
crowell: when we hit num entries == num buckets, this is the cannonical way to do hashtable growth.

src/ht.c
crowell: added to ls.h/c

src/ht.c
crowell: not any more

src/ht.c
crowell: not any more.

src/ht.c
crowell: now uses the sizes used in c#'s hashtable impl.

src/ht.c
radare: Cant we just keep those buckets to null until they need to be added to reduce memory and initial cpu usage

src/ht.c
radare: C# ftw

radare:
can you squash all those commits to ease the review? thx!

src/disk.c
```diff
@@ -57,7 +57,9 @@ SDB_API int sdb_disk_create (Sdb* s) {
 	memcpy (str, s->dir, nlen+1);
 	r_sys_mkdirp (str);
 	memcpy (str+nlen, ".tmp", 5);
-	close (s->fdump);
+	if (s->fdump != -1) {
```
radare: i would merge this in a separate PR maybe not related to hashtable

radare: and reasonable :)

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
- * authors.
- *
- * Authors:
- *    Eric Anholt <eric@anholt.net>
- *    Keith Packard <keithp@keithp.com>
- * Integration in r2 core api and hackit up
- *    pancake <nopcode.org>
- */
-
 #include "ht.h"
```
radare: add copyright note

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
+// hashfunction - the function that does the hashing, must not be null.
+// comparator - the function to check if values are equal, if NULL, just checks == (for storing ints).
+// keydup - function to duplicate to key (eg strdup), if NULL just does =.
+// valdup - same as keydup, but for values
+// pair_free - function for freeing a keyvaluepair - if NULL just does free.
+// calcsize - functino to calculate the size of a value. if NULL, just stores 0.
+SdbHash* internal_ht_new(ut32 size, HashFunction hashfunction, ListComparator comparator, DupKey keydup, DupValue valdup, HtKvFreeFunc pair_free, CalcSize calcsize) {
+	SdbHash* ht;
+	ut32 i;
+	ht = calloc (1, sizeof (*ht));
```
radare: initialize at the same line of deifning it

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
+	if (ht) {
```
radare: probably better to just if(!ht) { return NULL; }  instead of a big block

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
+		ht->count = 0;
+		ht->prime_idx = 0;
+		ht->load_factor = 1;
+		ht->hashfn = hashfunction;
+		ht->cmp = comparator;
+		ht->dupkey = keydup;
+		ht->dupvalue = valdup;
+		ht->table = calloc (ht->size, sizeof (SdbList*));
+		ht->calcsize = calcsize;
+		if (ht->table) {
```
radare: do not initialize all items here, we should initialize the list only when it is necessary, otherwise NULL is better than r_list_empty()

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
+	SdbKv* kvp = NULL;
+	ut32 computed_hash;
+	if (!hash) {
+		computed_hash = ht->hashfn (key);
+	} else {
+		computed_hash = *hash;
+	}
+	bucket = computed_hash % ht->size;
+	list = ht->table[bucket];
+	ls_foreach (list, iter, kvp) {
```
radare: iirc ls_foreach already checks for null, sot he previous initializer statement will not be necesary

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
-	ht->rehash = hash_sizes[ht->size_index].rehash;
-	ht->max_entries = hash_sizes[ht->size_index].max_entries;
-	ht->entries = 0;
-	ht->deleted_entries = 0;
-	for (e = old_ht.table; e != old_ht.table + old_ht.size; e++) {
-		if (entry_is_present (e))
-			ht_insert (ht, e->hash, e->data, e->iter);
+void ht_free(SdbHash* ht) {
+	ut32 i;
+	for (i = 0; i < ht->size; ++i) {
```
radare: do not preincrement if it is not necessary

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
+			if (ht->calcsize) {
+				kvp->value_len = (ut32)ht->calcsize (kvp->value);
+			} else {
+				kvp->value_len = 0;
+			}
+			ls_prepend (ht->table[bucket], kvp);
+			ht->count++;
+			// Check if we need to grow the table.
+			if (ht->count >= ht->load_factor * ht_primes_sizes[ht->prime_idx]) {
+				if (ht->prime_idx < sizeof (ht_primes_sizes)/sizeof (ht_primes_sizes[0])) {
```
radare: missing space around /

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
+				kvp->value = (char*)value;
+			}
+			bucket = hash % ht->size;
+			kvp->expire = 0;
+			if (ht->calcsize) {
+				kvp->value_len = (ut32)ht->calcsize (kvp->value);
+			} else {
+				kvp->value_len = 0;
+			}
+			ls_prepend (ht->table[bucket], kvp);
```
radare: what if inserting the key twice? will the list grow?

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
-			return 1;
+bool ht_insert_kvp(SdbHash* ht, SdbKv* kvp, bool update) {
+	bool found;
+	ut32 bucket, hash = ht->hashfn (kvp->key);
+	if (update) {
+		(void)ht_delete_internal (ht, kvp->key, &hash);
+	} else {
+		(void)ht_find (ht, kvp->key, &found);
+	}
+	if (update || !found) {
```
radare: merge those two conditionals into a single statemnt (if (kvp && (update || !found)) {

src/ht.c
```diff
@@ -1,239 +1,238 @@
...
-			entry->hash = hash;
-			entry->data = data;
-			entry->iter = rehash? iter: ls_append (ht->list, data);
-			ht->entries++;
-			return 1;
+bool ht_insert_kvp(SdbHash* ht, SdbKv* kvp, bool update) {
+	bool found;
+	ut32 bucket, hash = ht->hashfn (kvp->key);
+	if (update) {
+		(void)ht_delete_internal (ht, kvp->key, &hash);
```
radare: why is hash passed as reference?

src/sdb.c
```diff
@@ -175,7 +175,7 @@ SDB_API const char *sdb_const_get_len (Sdb* s, const char *key, int *vlen, ut32
 	if (s->fd == -1)
 		return NULL;
 	(void) cdb_findstart (&s->db);
-	if (cdb_findnext (&s->db, hash, key, keylen) < 1)
+	if (cdb_findnext (&s->db, sdb_hash (key), key, keylen) < 1)
```
radare: what if sdb_hash != hashtable hash thing? imho sdb should only permit 1 string hashing method and be consistent with it everywhere

src/sdb.c
```diff
@@ -382,29 +382,26 @@ static int sdb_set_internal (Sdb* s, const char *key, char *val, int owned, ut32
...
 			if (cas && kv->cas != cas)
 				return 0;
 			kv->cas = cas = nextcas ();
 			if (owned) {
 				kv->value_len = vlen;
-				free (kv->value);
 				kv->value = val; // owned
 			} else {
-				if (vlen>kv->value_len) {
+				if ((ut32)vlen>kv->value_len) {
```
radare: missing space around >

src/sdb.c
```diff
@@ -421,7 +418,7 @@ static int sdb_set_internal (Sdb* s, const char *key, char *val, int owned, ut32
 	}
 	if (kv) {
 		kv->cas = nextcas ();
-		ht_insert (s->ht, hash, kv, NULL);
+		ht_insert_kvp (s->ht, kv, 1 /*update*/);
```
radare: boolify maybe?

src/sdb.c
```diff
@@ -382,29 +382,26 @@ static int sdb_set_internal (Sdb* s, const char *key, char *val, int owned, ut32
...
+				if ((ut32)vlen>kv->value_len) {
 					free (kv->value);
 					kv->value = strdup (val);
 				} else {
 					memcpy (kv->value, val, vlen);
 				}
 				kv->value_len = vlen;
 			}
-		} else ht_delete_entry (s->ht, e);
+		} else ht_delete (s->ht, key);
```
radare: add braces

src/sdb.c
```diff
@@ -454,58 +451,69 @@ SDB_API SdbList *sdb_foreach_list (Sdb* s) {
...
+	ut32 i;
 	if (!s || !s->ht)
 		return;
-	ls_foreach (s->ht->list, iter, kv) {
-		if (!kv->value || !*kv->value)
-			continue;
-		printf ("%s=%s\n", kv->key, kv->value);
+	for (i = 0; i < s->ht->size; ++i) {
+		ls_foreach (s->ht->table[i], iter, kv) {
+			if (!kv->value || !*kv->value)
```
radare: moar braces

radare: maybe just invert the conditional to avoid having a continue in there? (if kv->value && *kv->value)

src/sdb.c
```diff
@@ -717,9 +723,9 @@ SDB_API int sdb_expire_set(Sdb* s, const char *key, ut64 expire, ut32 cas) {
 
 SDB_API ut64 sdb_expire_get(Sdb* s, const char *key, ut32 *cas) {
 	SdbKv *kv;
-	ut32 hash = sdb_hash (key);
-	kv = (SdbKv*)ht_lookup (s->ht, hash);
-	if (kv && *kv->value) {
+	bool found;
```
radare: initialize to false, if ht_find_kvp have a bug (didnt chkd src) it may fail.. but... why do we need a &found when the function is returning the result which must be != NULL in case of being valid?

src/ht.c
crowell: nope, list wont grow, if inserting the second time, we hit the `ht_delete_internal`

src/ht.c
crowell: because if it is `NULL` then we have to compute the hash ourself, otherwise, we take the value of the `*hash`.

src/sdb.c
crowell: this is a `->` operator, not a `>` operator.

src/sdb.c
crowell: this is because if you want to store intptr_t as the values you may want to be able to return NULL. this allows you to store NULL in the ht as a value.

src/sdb.c
radare: there's also a > :P

radare:
stress1 make[1]: *** [stress1] Segmentation fault (core dumped)

radare:
squash rebase and make it optional at compile time

src/sdb.c
```diff
@@ -208,25 +208,19 @@ SDB_API int sdb_unset (Sdb* s, const char *key, ut32 cas) {
...
 SDB_API int sdb_remove(Sdb *s, const char *key, ut32 cas) {
-	SdbHashEntry *e;
-	ut32 hash = sdb_hash (key);
-	e = ht_search (s->ht, hash);
-	if (e) {
-		ht_delete_entry (s->ht, e);
-		ls_delete (s->ht->list, e->iter);
-		return 1;
-	}
+	ht_delete (s->ht, key);
```
radare: this is wrong, should return 1 in case the key doesnt' exist

src/sdb.c
```diff
@@ -382,29 +382,28 @@ static int sdb_set_internal (Sdb* s, const char *key, char *val, int owned, ut32
...
 			if (cas && kv->cas != cas)
 				return 0;
 			kv->cas = cas = nextcas ();
 			if (owned) {
 				kv->value_len = vlen;
-				free (kv->value);
 				kv->value = val; // owned
 			} else {
-				if (vlen>kv->value_len) {
+				if ((ut32)vlen > kv->value_len) {
```
radare: why not just make vlen ut32 and drop the cast?

src/sdb.c
```diff
@@ -454,58 +453,69 @@ SDB_API SdbList *sdb_foreach_list (Sdb* s) {
...
-	ls_foreach (s->ht->list, iter, kv) {
-		if (!kv->value || !*kv->value)
-			continue;
-		if (!cb (user, kv->key, kv->value))
-			return 0;
+	for (i = 0; i < s->ht->size; ++i) {
+		ls_foreach (s->ht->table[i], iter, kv) {
+			if (!kv->value || !*kv->value) {
+				continue;
+			} else if (!cb (user, kv->key, kv->value)) {
```
radare: else is not necessary because the previous block breaks the flow

src/ht.c
```diff
@@ -1,239 +1,236 @@
...
-	{ 536870912,	590559793,	590559791 },
-	{ 1073741824,	1181116273,	1181116271},
-	{ 2147483648ul,	2362232233ul,	2362232231ul}
-};
+bool ht_delete_internal(SdbHash* ht, const char* key, ut32* hash) {
+	ut32 bucket;
+	SdbListIter* iter = NULL;
+	SdbList* list = NULL;
+	SdbKv* kvp = NULL;
+	ut32 computed_hash;
```
radare: use terminary operator

src/ht.c
```diff
@@ -1,239 +1,236 @@
...
-	{ 67108864,	73819861,	73819859  },
-	{ 134217728,	147639589,	147639587 },
-	{ 268435456,	295279081,	295279079 },
-	{ 536870912,	590559793,	590559791 },
-	{ 1073741824,	1181116273,	1181116271},
-	{ 2147483648ul,	2362232233ul,	2362232231ul}
-};
+bool ht_delete_internal(SdbHash* ht, const char* key, ut32* hash) {
+	ut32 bucket;
+	SdbListIter* iter = NULL;
```
radare: no need to initialize them

src/ht.c
```diff
@@ -1,239 +1,236 @@
...
+	ut32 computed_hash;
+	if (!hash) {
+		computed_hash = ht->hashfn (key);
+	} else {
+		computed_hash = *hash;
+	}
+	bucket = computed_hash % ht->size;
+	list = ht->table[bucket];
+	ls_foreach (list, iter, kvp) {
+		if (ht->cmp) {
```
radare: as long as the body is the same is better to just calculate an `is_equal` and separte comparations and logic

src/ht.c
```diff
@@ -1,239 +1,236 @@
...
 
-void ht_delete_entry(SdbHash *ht, SdbHashEntry *entry) {
-	if (!ht || !entry)
-		return;
-	if (!rehash && entry->iter) {
-		ls_delete (ht->list, entry->iter);
-		entry->iter = NULL;
+char* ht_find(SdbHash* ht, const char* key, bool* found) {
+	SdbKv* kvp = ht_find_kvp (ht, key, found);
+	if (*found) {
```
radare: use ternary operator. check if found is NULL

radare:
I have merged and fixed the comments in c6bb35f71e67aa7a3c2e59063bc4e6d032b14ebf

Thanks!

About speed, looks like there's no difference by running the tests/

PR #107, opened 2016-07-29
Handle Mingw-w64 and MSYS2

XVilka:


config.mk
```diff
@@ -92,6 +92,13 @@ ifeq (${OS},w32)
 OSTYPE=MINGW32
 endif
 
+ifneq (,$(findstring mingw32,${OSTYPE})$(findstring mingw64,${OSTYPE})$(findstring msys,${OSTYPE})$(findstring cygwin,${OSTYPE}))
```
radare: check for mingw only

PR #109, opened 2016-09-13
Correcting strcmp() in sdb_set_internal

tigertoes:
After my misunderstanding in radare/radare2#5739, I've checked out sdb and noticed that [test/cas.c](https://github.com/radare/sdb/blob/master/test/cas.c) appears to fail on a fresh checkout. I am assuming the tests are correct and that the functionality of setting the same key twice should not fail.

Before:

    time ./cas
    r2 = c2
    error
    r0 = c2
    failed
            0.00 real         0.00 user         0.00 sys

After this change is applied:

    cc -o cas cas.c ../src/libsdb.a -I../../src -I ../src
    time ./cas
    r2 = c2
    r3 = c3
            0.00 real         0.00 user         0.00 sys

### DIFF
--- a/test/cas.c
+++ b/test/cas.c
@@ -7,14 +7,24 @@ int main(int argc, char **argv) {
 	r = sdb_set (s, "hello", "world", 1);
 	sdb_const_get (s, "hello", &cas);
 	printf ("r%d = c%u\n", r, cas);
-	if (r != cas) printf ("error\n");
+	if (r != cas) {
+		printf ("error\n");
+		return 1;
+	}
 
 	r = sdb_set (s, "hello", "world", r);
 	sdb_const_get (s, "hello", &cas);
-	if (r != cas) printf ("error\n");
+	if (r != cas) {
+		printf ("error\n");
+		return 1;
+	}
 
 	printf ("r%d = c%u\n", r, cas);
-	if (r==0) printf ("failed\n");
+	if (r==0) {
+		printf ("failed\n");
+		return 1;
+	}
+
 	sdb_free (s);
 	return 0;
 }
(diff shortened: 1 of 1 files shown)


src/sdb.c
```diff
@@ -449,7 +449,7 @@ static int sdb_set_internal (Sdb* s, const char *key, char *val, int owned, ut32
 			if (cas && kv->cas != cas) {
 				return 0;
 			}
-			if (vlen == kv->value_len && !strcmp (kv->value, val)) {
+			if (vlen == kv->value_len && strcmp (kv->value, val) != 0) {
```
radare: this is wrong. the reason for this check is to ensure that the value is the same, your change swaps the condition so it will not update the value in case its different

tigertoes: Thanks for your feedback, would this mean that the tests in cas.c (which appear to have also failed [in the last successful build on the CI](https://travis-ci.org/radare/sdb/builds/152275982#L279)) are incorrect, not the code itself?

radare: but yep. i have noticed that the test is not passing and for some reason is not failing on travis. thats why i didnt noticed that :(

tigertoes: Okay, would you like me to revert this change and use this pull request to fix the failing tests? In addition, I think fixing the tests to exit 1 would be a good idea to detect this earlier.

radare:
yeah the test looks good now. but not the behaviour. thanks for reporting, i would like to merge when this test passes.

tigertoes:
Thank you, I intend on working towards fixing this later this evening. Your feedback would be appreciated.

trufae:
Fixed in be217e3f68c8798cb75a3b5315620497ec887a8b. Thanks for pointing out!

PR #111, opened 2016-10-05
Fix value always equals to key issue

ret2libc:
I started adding some unit testing for SDB.... first test, and it looks like at the moment the whole thing should be completely broken :( :( Indeed key and value are always the same when inserted through `ht_hash_insert`. 

What I would like to do is to make this HT implementation a little bit more generic, so that we can use that in r2 easily for other needs and not only for string-string pairs.

### DIFF
--- a/src/ht.c
+++ b/src/ht.c
@@ -30,7 +30,7 @@ const int ht_primes_sizes[] = {
 // keydup - function to duplicate to key (eg strdup), if NULL just does =.
 // valdup - same as keydup, but for values
 // pair_free - function for freeing a keyvaluepair - if NULL just does free.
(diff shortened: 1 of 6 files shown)


radare:
Yes im not very happy with the new hastable but i wasnt either of the old. The whole ht needs another rewrite, i did some benchmarka about memory usage and performance that would be good to run while reworking this

ret2libc:
I think we don't need a rewrite. It should work IMO, the implementation seems good and generic enough to be reused also somewhere else. Maybe there are just some small bugs that need to be fixed :) that's why i'm doing the tests. but, personally, i like it a lot :)

PR #128, opened 2017-03-07
Fixed some warns on windows

wargio:


src/sdb.h
```diff
@@ -32,8 +32,10 @@ extern "C" {
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <io.h>
-extern void *__cdecl _aligned_malloc(size_t, size_t);
+extern __attribute__((dllimport)) void *__cdecl _aligned_malloc(size_t, size_t);
 extern char *strdup (const char *);
+#undef r_offsetof
+#define r_offsetof(type, member) ((unsigned long) (ut64)&((type*)0)->member)
```
radare: I dont think r_offsetof is needed in sdb

wargio: well, it's defined there.. (see: https://github.com/wargio/sdb/blob/51bed648dfa78b887a00db14c4d30d91f97193e8/src/sdb.h#L20)

radare: this must be removed, its not used anywhere and i dont see why the changelog shows that this is new

wargio: No. that definition is at line 37, the other one is at line 20.
i've just changed it due the cast warning (from 8 bytes to 4 bytes)

wargio: and it's used in the code.
if removed, what is the correct code to be used then?

radare: i dont understand why you wanna cast a pointer to a 64bit number and then wrap it to 32bit to just get the delta between a type and the field? also, maybe im blind but im grepping for offsetof and i see 0 results

wargio: hmmm.. wait. i have to check. i'm confused.

PR #138, opened 2017-05-29
fixed some files to compile with msvc

skuater:
- avoid linkage missing functions

### DIFF
--- a/src/cdb.c
+++ b/src/cdb.c
@@ -101,7 +101,7 @@ bool cdb_read(struct cdb *c, char *buf, ut32 len, ut32 pos) {
 		return false;
 	}
 	while (len > 0) {
-		ssize_t r = read (c->fd, buf, len);
+		ut32 r = read (c->fd, buf, len);
 		if (r < 1 || (ut32) r != len) {
 			return false;
 		}
--- a/src/cdb.c
+++ b/src/cdb.c
@@ -101,7 +101,7 @@ bool cdb_read(struct cdb *c, char *buf, ut32 len, ut32 pos) {
 		return false;
 	}
 	while (len > 0) {
-		ut32 r = read (c->fd, buf, len);
+		int r = read (c->fd, buf, len);
 		if (r < 1 || (ut32) r != len) {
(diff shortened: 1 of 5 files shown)


src/sdb.h
```diff
@@ -37,8 +37,13 @@ extern "C" {
 #include <sys/stat.h>
 #include <io.h>
 #ifndef _MSC_VER
-extern __attribute__((dllimport)) void *__cdecl _aligned_malloc(size_t, size_t);
+extern __attribute__ ((dllimport)) void *__cdecl _aligned_malloc (size_t, size_t);
```
radare: nope, spaces before ( only are used when calling a function not in definition

src/cdb.c
```diff
@@ -101,7 +101,7 @@ bool cdb_read(struct cdb *c, char *buf, ut32 len, ut32 pos) {
 		return false;
 	}
 	while (len > 0) {
-		ssize_t r = read (c->fd, buf, len);
+		ut32 r = read (c->fd, buf, len);
```
radare: this is wrong. it will fail if -1

src/cdb.c
```diff
@@ -101,7 +101,7 @@ bool cdb_read(struct cdb *c, char *buf, ut32 len, ut32 pos) {
 		return false;
 	}
 	while (len > 0) {
-		ssize_t r = read (c->fd, buf, len);
+		int r = read (c->fd, buf, len);
```
radare: i think that if we cast to int it will be good on all platforms: `int r = (int)read ..`

src/sdb.h
```diff
@@ -37,8 +37,13 @@ extern "C" {
 #include <sys/stat.h>
 #include <io.h>
 #ifndef _MSC_VER
-extern __attribute__((dllimport)) void *__cdecl _aligned_malloc(size_t, size_t);
+extern __attribute__((dllimport)) void *__cdecl _aligned_malloc (size_t, size_t);
```
radare: ehm the space

PR #141, opened 2017-06-17
Fix crash on windows

xarkes:


src/sdb.c
```diff
@@ -551,7 +551,12 @@ static ut32 sdb_set_internal(Sdb* s, const char *key, char *val, int owned, ut32
 		return 0;
 	}
 	if (!val) {
-		val = "";
+		if (owned) {
+			val = (char*) malloc (1);
```
radare: What about using calloc(1.,1) so you dint have to initialize the first byte.. or maybe just strdup("")

PR #154, opened 2018-07-31
meson.build: bump version

ret2libc:


ret2libc:
A bit more explanation in the commits.

ret2libc:
do not merge yet, still working on it

ret2libc:
I think this looks good enough now :)

ret2libc:
Actually before merging I want to check a thing

ret2libc:
Ok I adapted a bit the things after thinking more about this and asking for help from others.

When sdb will be installed standalone (it's still not possible with meson, but i plan to make it possible soon), headers will be installed in `/usr/include/sdb`. When instead the meson.build is called as a "subproject", headers are not installed and we leave the choice to whoever is bundling sdb to choose where to installed it (e.g. radare2 will install the headers in `/usr/include/libr/sdb`).

ret2libc:
Btw, this is now ready for review @radare

ret2libc:
Closing it for now since I won't have time to see if anything bad happens because I'll go on vacation soon.

PR #158, opened 2018-09-20
create pkg-config file in meson and make tests runnable from meson

ret2libc:


test/Makefile
```diff
@@ -4,8 +4,8 @@ BINS=siolpain fmt stack cas hook hook2 stress1 fmtarr
 BINS+=bench-expire expire dumptwice syncget nsabuse drain merge
 BINS+=chkkey
 
-CFLAGS+=-I../../src -I ../src ${USER_CFLAGS}
-LDFLAGS+=../src/libsdb.a ${USER_LDFLAGS}
+CFLAGS+=$(shell pkg-config --cflags sdb) ${USER_CFLAGS}
```
radare: this builds against system wide sdb, not the one in current builddir, i understand that in meson the buildding is not the srcdir, but with this change, we are forced to use pkg-config and install it in the system to work.

i would prefer to be able to `make; make test; make install` rather than swaping the last 2 actions

ret2libc:
So I have considered this fact and the reasons I went ahead with this anyway were:
1. radare2 does the same
2. without pkg-config you are forced to have all those "../../../" all around CFLAGS/LDFLAGS
3. make builds in the src dir, so you can test only one version of sdb at a given time, otherwise you have to recompile everything. Meson can build in directories with different names, so you can't assume that all the stuff will be in `../../build/`
4. with pkg-config you are not forced to install libs/headers system-wide but it's enough to correctly set PKG_CONFIG_PATH/LD_LIBRARY_PATH/PATH like we do for radare2 travis and you can easily switch between different installation types
5. you test what the final user will actually use.

I understand it is handy to just do `make; make tests` but without pkg-config it becomes a big mess and very much "path dependent".

ret2libc:
For example, I just found out that the acr+make install wasn't working well because it didn't install some required headers. This was not caught by tests before because they were just using the source directory.

radare:
which required headers? sdb builds fine in travis without installing anything

ret2libc:
As said, it was running well because you are not installing it. If a user would try to install sdb and use it from the system, it wouldn't work, because `src/sdbht.h` was not installed on the system (see the changes in this PR)

ret2libc:
@radare ping

ret2libc:
What problem of static/dynamic linking?

radare:
I prefer static linking because this makes sdb self contained and give no room to problems like envvars or different versions of libs mainly because libsdb is not versioned at all and will cause major on problems on many people if they have multiple versions installed. Thats why sdb is embedded inside r2 and why static linking is prefered in the current makefile.

Btw pkgconfig doesnt works much well when having static and dynamic options of the same lib. Its usually better to have two separate pc files instead. Which sucks but cant think of a better solution

ret2libc:
> because libsdb is not versioned at all and will cause major on problems on many people if they have multiple versions installed

That's why in my previous commit about meson I've versioned the dynamic library (https://github.com/radare/sdb/blob/master/meson.build#L61)

> Btw pkgconfig doesnt works much well when having static and dynamic options of the same lib. Its usually better to have two separate pc files instead. Which sucks but cant think of a better solution

Again, I don't think these problems apply with sdb, since the pkg-config file is very simple and sdb doesn't have dependencies. Indeed just a:
`cc --static $(pkg-config --cflags --static sdb) -o test_hash ./test_hash.c $(pkg-config --libs --static sdb)` works well.

ret2libc:
> im not convinced on having pkgcondig as dependency at compile time

I see some reasons why this is not really a problem.
1. sdb is used only in radare2 AFAIK and radare2 already requires pkg-config
2. pkg-config can be made optional (though that doesn't prevent tests in travis from requiring it)

ret2libc:
Moreover, if it becomes *really* necessary, we can create a `env.sh` script similar to the r2 one to switch between different "install" directories.

ret2libc:
@radare ping

ret2libc:
Ok I'm trying to keep tests runnable without installing sdb.

radare:
what about having sdb.mk or deps.mk and put all the logic of defining flags and path to SDB in only one place? just include it from the shellscripts and the makefiles

PR #159, opened 2018-09-23
test: api tests do not make the build fail, though they should

ret2libc:


ret2libc:
@radare as written on Telegram, there's a bug in the sdb_array_add_num api, so this is probably going to be red.

radare:
why not fix the bug before making this change?

ret2libc:
Sorry, i'm not that much into sdb neither that interested in it :( We can things as they were if you prefer.

radare:
should be good to merge now, please rebase

PR #160, opened 2018-09-24
SdbHt: reintroduce growing

ret2libc:


src/sdb.c
```diff
@@ -650,16 +610,16 @@ static int sdb_foreach_list_cb(void *user, const char *k, const char *v) {
 	SdbList *list = (SdbList *)user;
 	SdbKv *kv = R_NEW0 (SdbKv);
 	/* seems like some k/v are constructed in the stack and cant be used after returning */
-	kv->key = strdup (k);
-	kv->value = strdup (v);
+	kv->base.key = strdup (k);
```
condret: what happens if strdup fails?

radare:
there are build errors. not gonna review something that is not even building

```
cd api && make
make[2]: Entering directory `/home/travis/build/radare/sdb/test/api'
/tmp/cclxC58f.o: In function `main':
array.c:(.text+0x12d7): multiple definition of `main'
/tmp/ccrn7qoh.o:array.c:(.text+0x12d7): first defined here
collect2: error: ld returned 1 exit status
make[2]: *** [all] Error 1
make[2]: Leaving directory `/home/travis/build/radare/sdb/test/api'
make[1]: [all] Error 2 (ignored)
make[2]: Entering directory `/home/travis/build/radare/sdb/test/unit'
gcc -I../../src -g -O0 test_array.c -o test_array ../../src/libsdb.a
In file included from test_array.c:1:0:
minunit.h:25:23: fatal error: sdb/types.h: No such file or directory
 #include <sdb/types.h>
                       ^
compilation terminated.
make[2]: *** [test_array] Error 1
make[2]: Leaving directory `/home/travis/build/radare/sdb/test/unit'
make[1]: *** [all] Error 2
make[1]: Leaving directory `/home/travis/build/radare/sdb/test'
make: *** [test] Error 2


```

src/ht.c
```diff
@@ -15,11 +17,30 @@ const int ht_primes_sizes[] = {
 	389357, 467237, 560689, 672827, 807403, 968897, 1162687,
 	1395263, 1674319, 2009191, 2411033, 2893249, 3471899,
 	4166287, 4999559, 5999471, 7199369
-#else
-	1024,
-#endif
 };
 
+#define HASHFN(ht, k) ((ht)->hashfn ? (ht)->hashfn (k) : (ut32)(ut64)(k))
```
radare: i dont like having so many macros

src/ht.c
```diff
@@ -174,45 +179,30 @@ static bool internal_ht_insert_kv(SdbHt *ht, HtKv *kv, bool update) {
...
-		} else {
-			kv->value = (void *)value;
-		}
-		kv->key_len = ht->calcsizeK ((void *)kv->key);
-		if (ht->calcsizeV) {
-			kv->value_len = ht->calcsizeV ((void *)kv->value);
-		} else {
-			kv->value_len = 0;
-		}
+		kv->key = DUPKEY (ht, key);
```
radare: may be null and its not checked later

src/sdbht.c
```diff
@@ -11,11 +11,11 @@ static bool sdb_ht_internal_insert(SdbHt* ht, const char* key,
 	}
 	SdbKv* kvp = calloc (1, sizeof (SdbKv));
 	if (kvp) {
-		kvp->key = strdup ((void *)key);
-		kvp->value = strdup ((void *)value);
-		kvp->key_len = strlen ((void *)kvp->key);
+		kvp->base.key = strdup ((void *)key);
```
radare: more places where allocations can fail

test/api/Makefile
```diff
@@ -1,8 +1,8 @@
 all:
-	@$(CC) $(CFLAGS) -I../../src array.c ../../src/libsdb.a
-	@./a.out
-	@$(CC) $(CFLAGS) -I../../src refs.c ../../src/libsdb.a
-	@./a.out
+	@$(CC) -o array array.c $(CFLAGS) -I../../src array.c ../../src/libsdb.a
```
radare: what about using make targets for those instead of inlined cc calls?

test/unit/minunit.h
```diff
@@ -6,14 +6,14 @@
 #ifndef TERMCOLOR_H
 #define TERMCOLOR_H
 
-#define TRED "\x1b[31m"
-#define TGREEN "\x1b[32m"
-#define TYELLOW "\x1b[33m"
-#define TBLUE "\x1b[34m"
+#define TRED     "\x1b[31m"
```
radare: dont indent that. is this clang-format?

test/unit/minunit.h
```diff
@@ -22,111 +22,106 @@
...
-			mu_test_status = MU_TEST_UNBROKEN; \
-		}                                          \
+#include <sdb/types.h>
+
+void sprint_mem(char *out, ut8 *buf, size_t len) {
+	size_t i;
+	*out = '\0';
+	for (i = 0; i < len; i++) {
+		if (i > 0) {
+			sprintf(out + strlen(out), " ");
```
radare: missing space before (

test/unit/minunit.h
```diff
@@ -22,111 +22,106 @@
...
+		}
+		sprintf (out + strlen(out), "%02x", buf[i]);
```
radare: another missing space

test/unit/minunit.h
```diff
@@ -22,111 +22,106 @@
...
+	}
+}
+
+#define mu_assert(message, test) do { \
+		if (!(test)) { \
+						mu_fail(message); \
```
radare: another one

src/ht.c
ret2libc: Why? Do you prefer static functions? Or dup code? I did macros to be sure the code was inlined and because they are very small

ret2libc:
> there are build errors. not gonna review something that is not even building

it is building... as you can see the `make` exits with 0. It's just because this PR depends on https://github.com/radare/sdb/pull/158 to make tests work.

src/ht.c
```diff
@@ -3,9 +3,11 @@
 #include "ht.h"
 #include "sdb.h"
 
+#define LOAD_FACTOR 0.9
```
radare: 0.9? load_factor is an ut32, not a float. and i would prefer to not depend on fpu

radare: oh! i see you deleted the variable... so there's no assignment. but this makes the fpu play into the game, and i would prefer to not depend on it for performance reasons

src/ht.c
radare: the reason for those macros are mainly to avoid nullderefs, what about making it into a function that wraps the call instead? like this: 

```
static inline void *wrap(void *a, void *b) {
  return a? a(b): NULL;
}
```

and do `wrap(ht->dupkey, k)` instead of having to dereference the pointer twice. obvously a compiler can do good optimization in both use cases, but looks cleaner to me than messing all those pointer calls with macros

src/ht.c
```diff
@@ -15,11 +17,30 @@ const int ht_primes_sizes[] = {
...
+#define HASHFN(ht, k) ((ht)->hashfn ? (ht)->hashfn (k) : (ut32)(ut64)(k))
+#define BUCKET(ht, k) ((HASHFN (ht, k) % (ht)->size))
+#define DUPKEY(ht, k) ((ht)->dupkey ? (ht)->dupkey (k) : (void *)k)
+#define DUPVAL(ht, v) ((ht)->dupvalue ? (ht)->dupvalue (v) : v)
+#define CALCSIZEK(ht, k) ((ht)->calcsizeK ? (ht)->calcsizeK (k) : 0)
+#define CALCSIZEV(ht, v) ((ht)->calcsizeV ? (ht)->calcsizeV (v) : 0)
+#define FREEFN(ht, kv) do { if ((ht)->freefn) { (ht)->freefn (kv); } } while (0)
+// when possible, use the precomputed prime numbers which help with collisions,
+// otherwise, at least make the number odd with |1
+#define COMPUTE_SIZE(idx, sz) ((idx) != UT32_MAX ? ht_primes_sizes[idx] : ((sz) | 1))
```
radare: sz|1 ?

radare: computersize can be a separate function

src/ht.c
```diff
@@ -15,11 +17,30 @@ const int ht_primes_sizes[] = {
...
+
+static inline bool is_kv_equal(SdbHt *ht, const char *key, const ut32 key_len, const HtKv *kv) {
+	if (key_len != kv->key_len) {
+		return false;
+	}
+
+	bool res = key == kv->key;
```
radare: if key == kv->key return true

src/ht.c
```diff
@@ -106,47 +92,67 @@ SDB_API bool ht_delete_internal(SdbHt* ht, const char* key, ut32* hash) {
...
 
-SDB_API void ht_free(SdbHt* ht) {
-	if (ht) {
-		ut32 i;
-		for (i = 0; i < ht->size; i++) {
-			ls_free (ht->table[i]);
-		}
-		free (ht->table);
-		ls_free (ht->deleted);
-#if INSERTORDER
```
radare: im ok to kill this, its been disabled for a while already

src/ht.c
```diff
@@ -106,47 +92,67 @@ SDB_API bool ht_delete_internal(SdbHt* ht, const char* key, ut32* hash) {
...
+
+	while (i < S_ARRAY_SIZE (ht_primes_sizes) &&
+		ht_primes_sizes[i] * LOAD_FACTOR < initial_size) {
+		i++;
 	}
+	if (i == S_ARRAY_SIZE (ht_primes_sizes)) {
+		i = UT32_MAX;
+	}
+
+	sz = COMPUTE_SIZE (i, (ut32)(initial_size * (2 - LOAD_FACTOR)));
```
radare: define and assign in the same line

src/ht.c
```diff
@@ -106,47 +92,67 @@ SDB_API bool ht_delete_internal(SdbHt* ht, const char* key, ut32* hash) {
...
-		ls_free (ht->deleted);
-#if INSERTORDER
-		ls_free (ht->list);
-#endif
-		free (ht);
+SDB_API SdbHt* ht_new_size(ut32 initial_size, DupValue valdup, HtKvFreeFunc pair_free, CalcSize calcsizeV) {
+	ut32 sz;
+	int i = 0;
+
+	while (i < S_ARRAY_SIZE (ht_primes_sizes) &&
```
radare: catch teh value of S_ARRAY_SIZE once before the while and use it in the conditional below too

src/ht.c
radare: moving the logic of COMPUTE size in a separate function will make this cleaner

src/ht.c
```diff
@@ -106,47 +92,67 @@ SDB_API bool ht_delete_internal(SdbHt* ht, const char* key, ut32* hash) {
...
 	HtKv* kv;
-	SdbListIter* iter;
-	ut32 i, sz = ht_primes_sizes[ht->prime_idx];
-	ht2 = internal_ht_new (sz, ht->hashfn, ht->cmp, ht->dupkey,
-			ht->dupvalue, (HtKvFreeFunc)ht->freefn, ht->calcsize);
-	ht2->prime_idx = ht->prime_idx;
+	SdbListIter* iter, *tmp;
+	ut32 i, idx, sz;
+
+	idx = ht->prime_idx != UT32_MAX ? ht->prime_idx + 1 : UT32_MAX;
```
radare: define and assign in the same line

src/ht.c
```diff
@@ -106,47 +92,67 @@ SDB_API bool ht_delete_internal(SdbHt* ht, const char* key, ut32* hash) {
...
+	sz = COMPUTE_SIZE (idx, ht->size * 2);
```
radare: same here and there

src/ht.c
```diff
@@ -155,14 +161,13 @@ static void internal_ht_grow(SdbHt* ht) {
...
 }
-#endif
 
 static bool internal_ht_insert_kv(SdbHt *ht, HtKv *kv, bool update) {
 	bool found = false;
 	if (!ht || !kv) {
 		return false;
 	}
-	ut32 bucket, hash = ht->hashfn (kv->key);
+	ut32 bucket, hash = HASHFN (ht, kv->key);
```
radare: hashfn should be always non-null imho, because your macro us using the k as int which is wrong hashing method (address of var != contents of var)

src/sdb.c
ret2libc: the end of the world as we know it :D

src/ht.c
ret2libc: I can use `static inline` functions, but I can't wrap all of them in just one `wrap` function, because some of them by default return 0, other by default just return the argument, etc.

src/ht.c
ret2libc: Ok. The `|1` is explained in the comment.
```
// when possible, use the precomputed prime numbers which help with collisions,
// otherwise, at least make the number odd with |1
```

src/ht.c
ret2libc: what do you mean? I don't get it.

src/ht.c
ret2libc: ok

src/ht.c
ret2libc: `k` is used directly only when a hashfn is not provided (e.g. to simulate a dict, since you said many times you think sometimes you just want numbers as keys). In general, I do believe that most of the times you actually want a hashfn, but I see no reasons to prevent numbers as keys.

There's a small catch here about `sizeof (void *) != sizeof (ut64)` on all archs, so in a future PR I will use `ut64` as types for key/value instead of `void*` as done in `dict.` What do you think?

src/ht.c
ret2libc: true.. it was never checked, but I can do it now.

test/api/Makefile
ret2libc: this will be fixed once I will rebase on top of PR #158

radare:
wat nao

ret2libc:
I will rebase this soon!

ret2libc:
Performance test:
```
#include "minunit.h"
#include <stdlib.h>
#include <sdb.h>
#include <time.h>

#define SEED 0xdeadbeef
#define INSERT_SZ   10000
#define FIND_SZ     1000000

#define INSERT_SZ2  10000000
#define FIND_SZ2    100000000

#define MOD_VAL    100000

void free_key(HtKv *kv) {
	free (kv->key);
	free (kv);
}

int main(int argc, char **argv) {
	// on master this defaults to allocating 1024 elements
	// on my PR it starts with 3 elements
	SdbHt *ht = ht_new (NULL, free_key, NULL);
	int i;

	clock_t start, end;
	double cpu_time_used, time_taken;

	// few elements
	srand(SEED);
	start = clock();
	for (i = 0; i < INSERT_SZ; ++i) {
		int val = rand() % MOD_VAL;
		char buf[20];
		snprintf (buf, 20, "%d", val);
		ht_insert (ht, buf, (void *)(ut64)val);
	}
	end = clock();
	time_taken = ((double)end - start)/CLOCKS_PER_SEC; // in seconds 
	printf("insert time: %f seconds\n", time_taken);

	long long sum = 0;
	start = clock();
	for (i = 0; i < FIND_SZ; ++i) {
		int val = rand() % MOD_VAL;
		char buf[20];
		snprintf (buf, 20, "%d", val);
		int res = (int)(ut64)ht_find (ht, buf, NULL);
		sum += res;
	}
	end = clock();
	time_taken = ((double)end - start)/CLOCKS_PER_SEC; // in seconds 
	printf("find time: %f seconds, sum = %lld\n", time_taken, sum);
	ht_free (ht);

	// many elements
	ht = ht_new (NULL, free_key, NULL);
	srand(SEED);
	start = clock();
	for (i = 0; i < INSERT_SZ2; ++i) {
		int val = rand() % MOD_VAL;
		char buf[20];
		snprintf (buf, 20, "%d", val);
		ht_insert (ht, buf, (void *)(ut64)val);
	}
	end = clock();
	time_taken = ((double)end - start)/CLOCKS_PER_SEC; // in seconds 
	printf("insert time (2 time): %f seconds\n", time_taken);

	sum = 0;
	start = clock();
	for (i = 0; i < FIND_SZ2; ++i) {
		int val = rand() % MOD_VAL;
		char buf[20];
		snprintf (buf, 20, "%d", val);
		int res = (int)(ut64)ht_find (ht, buf, NULL);
		sum += res;
	}
	end = clock();
	time_taken = ((double)end - start)/CLOCKS_PER_SEC; // in seconds 
	printf("find time (2 time): %f seconds, sum = %lld\n", time_taken, sum);

	ht_free (ht);
	return 0;
}
```

Results on master:
```
BASEDIR=$(pwd)/../../build-master make test_hash_performance ; ./test_hash_performance                                                         
cc -I/home/rschiron/projects/sdb/test/../src -I/home/rschiron/projects/sdb/test/unit/../../build-master  -g -O0 test_hash_performance.c -o test_hash_performance /home/rschiron/projects/sdb/test/unit/../../build-master/libsdb.a
insert time: 0.004317 seconds
find time: 0.413766 seconds, sum = 4738954741
insert time (2 time): 36.084946 seconds
find time (2 time): 348.878303 seconds, sum = 5000114468030
```

Results on my branch:
```
BASEDIR=$(pwd)/../../build make test_hash_performance ; ./test_hash_performance                                                                
cc -I/home/rschiron/projects/sdb/test/../src -I/home/rschiron/projects/sdb/test/unit/../../build  -g -O0 test_hash_performance.c -o test_hash_performance /home/rschiron/projects/sdb/test/unit/../../build/libsdb.a
insert time: 0.013456 seconds
find time: 0.175061 seconds, sum = 4738954741
insert time (2 time): 5.590546 seconds
find time (2 time): 43.648521 seconds, sum = 5000114468030
```

As you can see, this PR makes it just so much better in everything. And this is not even a fair comparison since the `ht` in my PR will start with 3 elements and not with 1024 as it is in master (which is many time a waste of space, as one of the issue has  shown)

ret2libc:
> The only waste is u allocate tons of sdb. Which is not the case. Fixed size is fine for most use cases

Yes, that's a problem that I'm going to fix in the next PR, where I will use an approach similar to `dict`. But for now, I think this PR is already good as it is, since it definitely improves over the previous implementation (which already used sdblists)

src/sdb.c
radare: thats not solving the issue :D

PR #162, opened 2018-10-02
Sdbkv refactor

ret2libc:
~~This PR is based on https://github.com/radare/sdb/pull/158 .~~

~~The only difference is the last commits.~~ The commit introduces helper macros in the sdbht.h file to access SdbKv->key/value/key_len/value_len since they are fields of the "base structure" HtKv. In another PR i'm going to change those fields, so I want to introduce these helpers to make those change easier.

### DIFF
--- a/src/query.c
+++ b/src/query.c
@@ -335,8 +335,7 @@ next_quote:
 			SdbListIter *iter;
 			SdbKv *kv;
 			ls_foreach (list, iter, kv) {
-				//eprintf ("(%s)(%s)\n", kv->key, kv->value);
-				foreach_list_cb (&user, kv->key, kv->value);
+				foreach_list_cb (&user, SDBKV_KEY (kv), SDBKV_VALUE (kv));
 			}
 			ls_free (list);
 			goto fail;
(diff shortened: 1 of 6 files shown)


ret2libc:
I intend to sync radare2 after merging this commit, so that in next PRs I can change
```
struct SdbKv {
	//sub of HtKv so we can cast safely
	char *key;
	char *value;
	ut32 key_len;
	ut32 value_len;
	ut32 cas;
	ut64 expire;
};
```
to
```
typedef struct sdb_kv {
	//sub of HtKv so we can cast safely
	HtKv base;
	ut32 cas;
	ut64 expire;
} SdbKv;
```
easily, without rewriting everything. Also, this will be helpful because key/value are both `char*` in SdbKv, but not in HtKv, so a cast will be needed.

src/sdbht.h
```diff
@@ -14,6 +14,11 @@ typedef struct sdb_kv {
 	ut64 expire;
 } SdbKv;
 
+#define SDBKV_KEY(kv) ((kv)->key)
```
ret2libc: @radare Do you prefer `static inline` functions instead of these macros?

src/query.c
```diff
@@ -335,8 +335,7 @@ SDB_API char *sdb_querys (Sdb *r, char *buf, size_t len, const char *_cmd) {
 			SdbListIter *iter;
 			SdbKv *kv;
 			ls_foreach (list, iter, kv) {
-				//eprintf ("(%s)(%s)\n", kv->key, kv->value);
-				foreach_list_cb (&user, kv->key, kv->value);
+				foreach_list_cb (&user, SDBKV_KEY (kv), SDBKV_VALUE (kv));
```
radare: why this change i dont see a reason to have this macro anymore

ret2libc:
> We probably dont need those macros at all

See this comment https://github.com/radare/sdb/pull/162#issuecomment-426960357

src/query.c
radare: i understand the concern of having macros to make this easier to refactor, but this is making the code harder to read, what about just having 1 letter macros for that? K(kv) and V(kv)

PR #164, opened 2018-10-05
Fix b64_* for Solaris

thestr4ng3r:
Solaris has its own b64_encode and b64_decode functions, which the ones here can conflict with:
```
Standard C Library Functions                                    b64_encode(3C)

NAME
       b64_encode,  b64_decode,  b32_encode,  b32_decode - encode and decode a
       string in base64 or base32

SYNOPSIS
       #include <string.h>

       ssize_t b64_encode(char restrict *outbuf, size_t outbufsz,
                const void restrict *inbuf, size_t inbufsz,
                const char *alphabet, uint64_t flags);

       ssize_t b64_decode(void *outbuf, size_t outbufsz,
                const char *inbuf, const char *alphabet, uint64_t flags);

```

### DIFF
--- a/src/base64.c
+++ b/src/base64.c
@@ -10,7 +10,7 @@
 static const char cb64[]="ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 static const char cd64[]="|$$$}rstuvwxyz{$$$$$$$>?@ABCDEFGHIJKLMNOPQRSTUVW$$$$$$XYZ[\\]^_`abcdefghijklmnopq";
 
-static void b64_encode(const ut8 in[3], char out[4], int len) {
+static void sdb_b64_encode(const ut8 in[3], char out[4], int len) {
 	if (len < 1) {
 		return;
 	}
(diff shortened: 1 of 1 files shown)


src/base64.c
```diff
@@ -10,7 +10,7 @@
 static const char cb64[]="ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/";
 static const char cd64[]="|$$$}rstuvwxyz{$$$$$$$>?@ABCDEFGHIJKLMNOPQRSTUVW$$$$$$XYZ[\\]^_`abcdefghijklmnopq";
 
-static void b64_encode(const ut8 in[3], char out[4], int len) {
+static void _b64_encode(const ut8 in[3], char out[4], int len) {
```
ret2libc: they told me it's better to avoid names that start with `_` or `__`

radare:
Lets just prefix it with r coz we own this letter

thestr4ng3r:
I used `sdb_`. Is that fine too?

radare:
Yeah sgtm

PR #166, opened 2018-10-09
sdb: add API to return (sorted) list filtered

ret2libc:
this will be useful in r2 cmd_type where a lot of times we iterate over all sdb just to get some keys that have a given prefix... https://github.com/radare/radare2/blob/master/libr/core/cmd_type.c#L335 https://github.com/radare/radare2/blob/master/libr/core/cmd_type.c#L320 etc. At the same time, we can return a sorted list, which will make the tests stable (because ht does not have ordering).

### DIFF
--- a/src/sdb.c
+++ b/src/sdb.c
@@ -616,8 +616,7 @@ static int sdb_foreach_list_cb(void *user, const char *k, const char *v) {
 }
 
 static int __cmp_asc(const void *a, const void *b) {
-	const SdbKv *ka = a;
-	const SdbKv *kb = b;
+	const SdbKv *ka = a, *kb = b;
 	return strcmp (sdbkv_key (ka), sdbkv_key (kb));
 }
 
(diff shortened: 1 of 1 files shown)


src/sdb.c
```diff
@@ -630,6 +630,36 @@ SDB_API SdbList *sdb_foreach_list(Sdb* s, bool sorted) {
...
+	SdbForeachCallback filter;
+	SdbList *list;
+};
+
+static int sdb_foreach_list_filter_cb(void *user, const char *k, const char *v) {
+	struct foreach_list_filter_t *u = (struct foreach_list_filter_t *)user;
+	SdbList *list = u->list;
+
+	if (!u->filter || u->filter (NULL, k, v)) {
+		SdbKv *kv = R_NEW0 (SdbKv);
```
radare: chk for null

src/sdb.c
```diff
@@ -630,6 +630,36 @@ SDB_API SdbList *sdb_foreach_list(Sdb* s, bool sorted) {
...
+		kv->base.key = strdup (k);
```
radare: if those strdups fail we shuoldnt add the element in the list

src/sdb.c
```diff
@@ -630,6 +630,36 @@ SDB_API SdbList *sdb_foreach_list(Sdb* s, bool sorted) {
...
+		kv->base.value = strdup (v);
+		ls_append (list, kv);
+	}
+	return 1;
+}
+
+SDB_API SdbList *sdb_foreach_list_filter(Sdb* s, SdbForeachCallback filter, bool sorted) {
+	SdbList *list = ls_newf ((SdbListFree)sdbkv_free);
+	struct foreach_list_filter_t u;
```
radare: allocation can fail

src/sdb.c
```diff
@@ -630,6 +629,50 @@ SDB_API SdbList *sdb_foreach_list(Sdb* s, bool sorted) {
...
+
+	if (!u->filter || u->filter (NULL, k, v)) {
+		kv = R_NEW0 (SdbKv);
+		if (!kv) {
+			goto err;
+		}
+		kv->base.key = strdup (k);
+		kv->base.value = strdup (v);
+		if (!kv->base.key || !kv->base.value) {
+			goto err;
```
radare: free (kv->base.key);
free (kv->base.value);

src/sdb.c
```diff
@@ -630,6 +629,50 @@ SDB_API SdbList *sdb_foreach_list(Sdb* s, bool sorted) {
...
+	return 1;
+ err:
+	sdbkv_free (kv);
+	return 0;
+}
+
+SDB_API SdbList *sdb_foreach_list_filter(Sdb* s, SdbForeachCallback filter, bool sorted) {
+	struct foreach_list_filter_t u;
+	SdbList *list = ls_newf ((SdbListFree)sdbkv_free);
+
```
radare: empty ilne

PR #168, opened 2018-10-11
Attempt to fix add_num test cases.

devnexen:


src/array.c
```diff
@@ -211,7 +211,14 @@ SDB_API int sdb_array_set_num(Sdb *s, const char *key, int idx, ut64 val,
 
 SDB_API int sdb_array_add_num(Sdb *s, const char *key, ut64 val, ut32 cas) {
 	char buf[SDB_NUM_BUFSZ];
-	char *v = sdb_itoa (val, buf, SDB_NUM_BASE);
+	int base = SDB_NUM_BASE;
+	char *f = sdb_array_get (s, key, 0, NULL);
+	if (f) {
+		// We assume the list contains same base...
```
radare: removing one assumption to add another assumption is probably not the best way to handle this

see that sdb_array_contains_num() should be used instead of sdb_array_add. but again, this will slow down a lot of performance, and this assumption is not going to fix all the possible cases.

PR #170, opened 2018-10-29
Windows&Unicode fix, fixed heap corruption

a1ext:
Related to radareorg/cutter#589
fixed heap corruption in SDB
fixed proper project saving on windows

### DIFF
--- a/src/disk.c
+++ b/src/disk.c
@@ -12,49 +12,34 @@
 #if __SDB_WINDOWS__
 
 #if UNICODE
-static wchar_t* r_str_mb_to_wc_l(const char *buf, int len) {
-	wchar_t *res_buf = NULL;
-	size_t sz;
-	bool fail = true;
 
-	if (!buf || len <= 0) {
+
+static wchar_t *r_utf8_to_utf16_l (const char *cstring, int len) {
+	if (!cstring || !len || len < -1) {
 		return NULL;
 	}
-	sz = mbstowcs (NULL, buf, len);
-	if (sz == (size_t)-1) {
-		goto err_r_str_mb_to_wc;
+	wchar_t *rutf16 = NULL;
+	int wcsize;
+
+	if ((wcsize = MultiByteToWideChar (CP_UTF8, 0, cstring, len, NULL, 0))) {
+		wcsize += 1;
+		if ((rutf16 = (wchar_t *) calloc (wcsize, sizeof (wchar_t)))) {
+			MultiByteToWideChar (CP_UTF8, 0, cstring, len, rutf16, wcsize);
+			if (len != -1) {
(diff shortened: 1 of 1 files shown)


src/disk.c
```diff
@@ -120,7 +105,13 @@ SDB_API bool sdb_disk_create(Sdb* s) {
 	if (s->fdump != -1) {
 		close (s->fdump);
 	}
+#if __SDB_WINDOWS__ && UNICODE
+	wchar_t *wstr = r_sys_conv_utf8_to_utf16 (str);
+	s->fdump = _wopen (wstr, O_BINARY | O_RDWR | O_CREAT | O_TRUNC, SDB_MODE);
```
radare: check for wstr

src/disk.c
```diff
@@ -12,49 +12,34 @@
...
 
-#define r_sys_conv_utf8_to_utf16(buf) r_str_mb_to_wc (buf)
+#define r_sys_conv_utf8_to_utf16(buf) r_utf8_to_utf16_l ((buf), -1)
 
 static bool r_sys_mkdir(char *path) {
 	LPTSTR path_ = r_sys_conv_utf8_to_utf16 (path);
 	bool ret = CreateDirectory (path_, NULL);
 
-	free (path);
+	free (path_);
```
radare: uhm, i think this function should be const char *path, and as long as the behaviour has changed, review the rest of the uses.. i did a quick view and seems like it's fine because the non-windows cases the string is still owned by the caller. Also, his non-processed string is saved in s->ndump, so we can probably just return the re-encoded string to save it in ndump and avoid to call r_sys_conv_utf8... again in sdb_disk_finish(). but this change can be probably done later in a separate pr

src/disk.c
```diff
@@ -12,49 +12,34 @@
...
+		}
 	}
-	return r_str_mb_to_wc_l (buf, strlen (buf));
+	return rutf16;
 }
 
-#define r_sys_conv_utf8_to_utf16(buf) r_str_mb_to_wc (buf)
+#define r_sys_conv_utf8_to_utf16(buf) r_utf8_to_utf16_l ((buf), -1)
 
 static bool r_sys_mkdir(char *path) {
```
radare: make it const char *

src/disk.c
a1ext: this will require to make `ndump` of `wchar_t *` type... which will require to put mess of `#if/#else` in `sdb.h`

a1ext: Also, I think it's better to make it impossible to build it on windows without turned on unicode support...

radare:
Thanks!

a1ext:
you are welcome

radare:
sgtm

PR #172, opened 2018-11-07
Commit number

radare:


radare:
@ret2libc wanna finish this change?

PR #174, opened 2018-11-07
ht: provide ht_new0 API for convenience

ret2libc:


src/ht.h
```diff
@@ -46,6 +46,7 @@ typedef struct ht_t {
 // If hashfunction is NULL it will be used sdb_hash internally
 // If keydup or valdup are null it will be used an assignment
 // If keySize or valueSize are null it will be used strlen internally
+SDB_API SdbHt* ht_new0();
```
radare: add void

src/ht.c
```diff
@@ -127,6 +127,14 @@ SDB_API SdbHt* ht_new(DupValue valdup, HtKvFreeFunc pair_free, CalcSize calcsize
 		valdup, pair_free, (CalcSize)strlen, calcsizeV, sizeof (HtKv));
 }
 
+static void free_kv_key(HtKv *kv) {
+	free (kv->key);
+}
+
+SDB_API SdbHt* ht_new0() {
```
radare: add void

src/ht.h
ret2libc: ah yeah! forgot that

PR #177, opened 2018-11-09
use "generics" approach to create multiple Ht types

ret2libc:
@radare if you want to rename things this seems like the right moment!


src/sdbht.h
```diff
@@ -1,12 +1,12 @@
...
 
-#include "ht.h"
+#include "ht_p.h"
 
 /** keyvalue pair **/
 typedef struct sdb_kv {
-	//sub of HtKv so we can cast safely
-	HtKv base;
+	//sub of HtPKv so we can cast safely
+	HtPKv base;
```
radare: HtPKv vs SdbPHt. looks pretty inconsistent to me, P is prefix, suffix, wtf

ret2libc:
@radare I fixed the naming! Please tell me what you think :)

src/ht_inc.c
```diff
@@ -0,0 +1,307 @@
...
+	5839, 7013, 8419, 10103, 12143, 14591, 17519, 21023,
+	25229, 30293, 36353, 43627, 52361, 62851, 75431, 90523,
+	108631, 130363, 156437, 187751, 225307, 270371, 324449,
+	389357, 467237, 560689, 672827, 807403, 968897, 1162687,
+	1395263, 1674319, 2009191, 2411033, 2893249, 3471899,
+	4166287, 4999559, 5999471, 7199369
+};
+
+static inline ut32 hashfn(HtName_(Ht) *ht, const KEY_TYPE k) {
+	return ht->opt.hashfn ? ht->opt.hashfn (k) : KEY_TO_HASH(k);
```
radare: missing space

test/bench.c
```diff
@@ -1,7 +1,7 @@
 #include <sdb-one.h>
 
 int main(int argc, char **argv) {
-	SdbHt *h = ht_new(free);
+	HtP *h = ht_new(free);
```
radare: space

src/sdbht.c
```diff
@@ -27,41 +27,41 @@ static bool sdb_ht_internal_insert(SdbHt* ht, const char* key,
...
 }
 
-SDB_API bool sdb_ht_insert_kvp(SdbHt* ht, SdbKv *kvp, bool update) {
-	return ht_insert_kv (ht, (HtKv*)kvp, update);
+SDB_API bool sdb_ht_insert_kvp(HtP* ht, SdbKv *kvp, bool update) {
+	return ht_p_insert_kv (ht, (HtPKv*)kvp, update);
 }
 
-SDB_API bool sdb_ht_update(SdbHt *ht, const char *key, const char*value) {
+SDB_API bool sdb_ht_update(HtP *ht, const char *key, const char*value) {
```
radare: missing space

ret2libc:
Yeah I couldn't actually think of other cases where you don't want a void* as value. The other naming could be:

`HtUU`, `HtUP`, `HtPP`. What about these?

radare:
Yeah sounds better to me. Not sure how many we will have to replicate but looks less confusing to me

ret2libc:
Ok, it's fine with me too. I think we won't need more then `HtUU`, `HtUP` and `HtPP`. They seem to cover pretty much all use cases to me. Of course, you can always subclass it even further, but with `void *` as type you cover everything quite easily.

I will rename things soon.

ret2libc:
Ok, Everything was addressed.

PR #200, opened 2019-10-24
Add sdb_copy()

thestr4ng3r:


thestr4ng3r:
ping @radare

src/sdb.h
```diff
@@ -122,6 +122,10 @@ SDB_API int sdb_count(Sdb* s);
 SDB_API void sdb_reset(Sdb* s);
 SDB_API void sdb_setup(Sdb* s, int options);
 SDB_API void sdb_drain(Sdb*, Sdb*);
+
+// Copy everything, including namespaces, from src to dst
+SDB_API void sdb_copy(Sdb *dst, Sdb *src);
```
radare: this operation is usually called clone on objects. the src sdb should be the first arg. otherwise it breaks the `db.copy(dst);` convention

PR #203, opened 2020-04-08
Remove a goto

eagleoflqj:
**Detailed description**

Step forward to fix LGTM alert "Multiple forward and backward goto statements".

**Test plan**

Make and make test, or white-box test.

**Closing issues**

...

### DIFF
--- a/src/query.c
+++ b/src/query.c
@@ -281,8 +281,7 @@ next_quote:
 		*next = 0;
 	}
 	arroba = strchr (cmd, '/');
-	if (arroba) {
-	next_arroba:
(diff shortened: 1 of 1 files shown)


src/query.c
```diff
@@ -281,8 +281,7 @@ SDB_API char *sdb_querys (Sdb *r, char *buf, size_t len, const char *_cmd) {
 		*next = 0;
 	}
 	arroba = strchr (cmd, '/');
```
radare, line 283: Rename this variable to slash instead of arroba

PR #207, opened 2020-04-09
Simplify code

eagleoflqj:
**Detailed description**

Not nessesary to decrease and increase quot.

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/src/query.c
+++ b/src/query.c
@@ -259,13 +259,10 @@ repeat:
 next_quote:
 		quot = strchr (quot, '"');
 		if (quot) {
-			quot--;
-			if (*quot=='\\') {
-				memmove (quot, quot + 1, strlen (quot));
-				quot++;
+			if (*(quot - 1) == '\\') {
+				memmove (quot - 1, quot, strlen (quot) + 1);
 				goto next_quote;
 			}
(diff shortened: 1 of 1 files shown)


src/query.c
```diff
@@ -259,13 +259,10 @@ SDB_API char *sdb_querys (Sdb *r, char *buf, size_t len, const char *_cmd) {
 next_quote:
 		quot = strchr (quot, '"');
 		if (quot) {
-			quot--;
-			if (*quot=='\\') {
-				memmove (quot, quot + 1, strlen (quot));
-				quot++;
+			if (*(quot - 1) == '\\') {
```
trufae, line 262: if string is "" that -1 can result on oobread.use a temporal variable to save the result of strchr and check if quot2 > quot before doing that

eagleoflqj: line 255 `*val` is already `"`, `quot` is at least that `val` plus one, and `quot = strchr (quot, '"');` doesn't decrease `quot` (if it doesn't return NULL), so `quot-1` is guaranteed legal.

ret2libc (approved):
LGTM

PR #209, opened 2020-07-02
Strip UTF8 BOM when creating db

kazarmy:
**Detailed description**

This pr prevents the following:

BOM_in_sdb_with_oval

i.e. a UTF8 BOM in an sdb, due to a UTF8 BOM that materialized in sdb.exe's input.

For this particular case, I'm not entirely sure how it materialized -- current hypothesis is that a combination of PowerShell's Get-Content + https://superuser.com/questions/269818/change-default-code-page-of-windows-console-to-utf-8/1435645#1435645 is the cause.

**Test plan**

Travis is green.

**Closing issues**

...

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -265,9 +265,13 @@ static int createdb(const char *f, const char **args, int nargs) {
 	insertkeys (s, args, nargs, '=');
 	sdb_config (s, options);
 	for (; (line = stdin_slurp (NULL));) {
-		if ((eq = strchr (line, '='))) {
+		size_t off = 0;
+		if (line[0] == '\xef' && line[1] == '\xbb' && line[2] == '\xbf') {
+			off = 3;
+		}
+		if ((eq = strchr (line + off, '='))) {
 			*eq++ = 0;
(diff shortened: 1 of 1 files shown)


kazarmy:
Edited the description.

src/main.c
```diff
@@ -265,9 +265,13 @@ static int createdb(const char *f, const char **args, int nargs) {
 	insertkeys (s, args, nargs, '=');
 	sdb_config (s, options);
 	for (; (line = stdin_slurp (NULL));) {
-		if ((eq = strchr (line, '='))) {
+		size_t off = 0;
+		if (line[0] == '\xef' && line[1] == '\xbb' && line[2] == '\xbf') {
```
trufae, line 269: Use strncmp

kazarmy:
> Is this affecting the very first line of every line?

No but just realized that today. Sorry.

> As long as its windows fault maybe worth check if theres a way to tell windows not to bom outputs maybe

Yeah see radareorg/radare2#17203. Closing this pr.

PR #210, opened 2020-07-08
Use _ instead of : because : is invalid on Windows

ret2libc:
The use of : in file names prevents sdb from being cloned on Windows

**Test plan**

Just try to clone this branch and master on a Windows machine. master will fail with something like:
```
error: invalid path 'test/fuzz/minimized_third/out/id:0000052.out'
```

**Closing issues**

None


trufae:
Lgtm

PR #215, opened 2020-09-19
Add sdb out.sdb == in1.sdb.txt in2.sdb.txt

thestr4ng3r:
For speeding upt the r2 build.
This can be later made to accept the new sdb text format with escaping and stuff.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -364,7 +364,7 @@ static int base64encode(void) {
 static int base64decode(void) {
 	ut8 *out;
 	size_t len, ret = 1;
-	char *in = (char *) slurp (stdin, &len);
+	char *in = slurp (stdin, &len);
 	if (in) {
 		int declen;
 		out = sdb_decode (in, &declen);
--- a/src/main.c
+++ b/src/main.c
@@ -34,12 +34,12 @@ static void write_null(void) {
 #define BS 128
 #define USE_SLURPIN 1
 
-static char *stdin_slurp(int *sz) {
+static char *file_slurp(FILE *f, int *sz) {
 	int blocksize = BS;
 	static int bufsize = BS;
 	static char *next = NULL;
-	static int nextlen = 0;
-	int len, rr, rr2;
+	static size_t nextlen = 0;
+	size_t len, rr, rr2;
 	char *tmp, *buf = NULL;
 	if (sz) {
 		*sz = 0;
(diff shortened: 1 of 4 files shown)


src/main.c
```diff
@@ -34,12 +34,12 @@ static void write_null(void) {
 #define BS 128
 #define USE_SLURPIN 1
 
-static char *stdin_slurp(int *sz) {
+static char *file_slurp(FILE *f, int *sz) {
```
trufae, line 37: as long as its taking a stream/file-handler as argument,its no longer reading from a file. so better rename this to `slurp`. also sz should be size_t

src/main.c
```diff
@@ -338,7 +364,7 @@ static int base64encode(void) {
 static int base64decode(void) {
 	ut8 *out;
 	int len, ret = 1;
-	char *in = (char *) stdin_slurp (&len);
+	char *in = (char *) file_slurp (stdin, &len);
```
trufae, line 367: i dont think the castis necessary here. also len must be size_t

src/main.c
```diff
@@ -501,6 +527,8 @@ int main(int argc, const char **argv) {
 		}
 	} else if (!strcmp (argv[db0 + 1], "=")) {
```
trufae, line 528: what about checking char by char instead of two consecurive strcmps?

trufae (changes requested):

src/main.c
thestr4ng3r: Instead of
```c
	} else if (*argv[db0 + 1] == '=') {
		ret = createdb (argv[db0], NULL, 0);
	} else if (!strcmp (argv[db0 + 1], "==")) {
		ret = createdb (argv[db0], argv + db0 + 2, argc - (db0 + 2));
	} else {
		s = sdb_new (NULL, argv[db0], 0);
		if (!s) {
			return 1;
		}
		...
```
you would then have something like:
```c
	} else if (*argv[db0 + 1] == '=') {
		if (argv[db0 + 1][1] == '\0') {
			ret = createdb (argv[db0], NULL, 0);
		} else if (argv[db0 + 1][1] == '=' && argv[db0 + 1][2] == '\0') {
			ret = createdb (argv[db0], argv + db0 + 2, argc - (db0 + 2));
		} else {
			goto els;
		}
	} else {
els:
		s = sdb_new (NULL, argv[db0], 0);
		if (!s) {
			return 1;
		}
		...
```
So I think the first is better.

trufae: No. This is ultracomplicated. You can nest one condition under the other. And make it shorter but anyway im in phone and strcmp is fine

PR #219, opened 2020-09-27
Add Plaintext SDB Format

### LINKED ISSUE #214: Support plaintext .sdb files
Identify when the .sdb file is k=v in plaintext instead of a disk hashtable in cdb format

This format should be defined somehow in the Sdb instance and saved in the same format when dumping.

thestr4ng3r:
see comments in the beginning of text.c for more info.

**Closing issues**

Fix  #214


src/text.c
```diff
@@ -0,0 +1,358 @@
...
+	if (!f) {
+		return false;
+	}
+	bool r = sdb_text_fsave (s, f, sort);
+	fclose (f);
+	return r;
+}
+
+SDB_API bool sdb_text_fload(Sdb *s, FILE *f) {
+	size_t bufsz = 16; // TODO: more
```
thestr4ng3r, line 201: Use more here

XVilka: 17?

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+	// sub-namespaces
+	l = s->ns;
+	if (sort) {
+		l = ls_clone (l);
+		ls_sort (l, cmp_ns);
+	}
+	SdbNs *ns;
+	ls_foreach (l, it, ns) {
+		fwrite ("\n", 1, 1, f);
+		ls_push (path, ns->name);
```
ret2libc, line 168: mmm is it ever possible to have more than one name in `path` being passed to `text_fsave`? It seems to me that this way `text_fsave` will always have at most one path to deal with, so maybe instead of passing a SdbList we can just pass a single string.

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+ *
+*/
+
+static int cmp_ns(const void *a, const void *b) {
+	const SdbNs *nsa = a;
+	const SdbNs *cia = b;
+	return strcmp (nsa->name, cia->name);
+}
+
+static bool text_fsave(Sdb *s, FILE *f, bool sort, SdbList *path) {
```
ret2libc, line 61: I think this whole function can be made much easier to follow by splitting stuff in functions (e.g. `write_ns_name` or something like that and then `write_entries` or something like that).

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+// p = position until we have already written everything
+// macro to flush a block of text that doesn't have to be escaped
+#define FLUSH do { if (p != n) { fwrite (p, 1, n - p, f); p = n; } } while (0)
+// macro to flush and skip a char
+#define SKIP do { FLUSH; p++; } while (0)
+	// path
+	fwrite ("/", 1, 1, f); // always print a /, even if path is empty
+	SdbListIter *it;
+	const char *path_token;
+	bool first = true;
```
ret2libc, line 72: this `first` thing can probably be avoided?

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+				if (buf[pos] == '/') {
+					state = STATE_PATH;
+					token_begin = 1;
+					pos++;
+					continue;
+				} else {
+					state = STATE_KEY;
+				}
+			}
+			if (buf[pos] == '\n') {
```
ret2libc, line 260: mmm i was expecting a sort of `switch` here, not a "random" check against `\n`. I think it would be much easier to follow if you use a switch on the state and then check the characters in the buffer according to the state you are in.

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+			bufsz = newsz;
+			char *newbuf = realloc (buf, bufsz);
+			if (!newbuf) {
+				ret = false;
+				break;
+			}
+			buf = newbuf;
+		}
+		// read full buffer (leave 1 for null-terminating)
+		size_t read = fread (buf + pos, 1, bufsz - 1 - pos, f);
```
ret2libc, line 236: use another name, `read` is already `read (2)` :D

ret2libc (commented):
I'm not super familiar with SDB internals, but by reading the issues it said SDB shall be able to automatically detect if loading a text format or a "compiled" one. Where is this thing done?

src/ls.c
```diff
@@ -260,6 +260,21 @@ SDB_API void *ls_pop(SdbList *list) {
 	return NULL;
 }
 
+SDB_API SdbList *ls_clone(SdbList *list) {
```
trufae, line 263: oh. this may conflict after merging the other pr

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+		const char *v = sdbkv_value (kv);
+		// escape leading '/'
+		if (*k == '/') {
+			fwrite ("\\", 1, 1, f);
+		}
+
+		// write and escape key
+		const char *p = k;
+		const char *n = p;
+		while (*n) {
```
trufae, line 117: doesnt seems to handle LF/NL dos newlines

src/text.c
trufae: #define more 16

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+		}
+		// read full buffer (leave 1 for null-terminating)
+		size_t read = fread (buf + pos, 1, bufsz - 1 - pos, f);
+		if (!read && !feof (f)) {
+			// error
+			ret = false;
+			break;
+		}
+		// process buffer
+		size_t buf_filled = pos + read;
```
trufae, line 243: read is a symbol. dont use varnames that conflict

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+		while (pos < buf_filled) {
+			if (state == STATE_NEWLINE) {
+				// at the start of a line, decide whether it's a path or a k=v
+				// by whether there is a leading slash.
+				if (buf[pos] == '/') {
+					state = STATE_PATH;
+					token_begin = 1;
+					pos++;
+					continue;
+				} else {
```
trufae, line 256: dont else after continue

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+					state = STATE_KEY;
+				}
+			}
+			if (buf[pos] == '\n') {
+				unescape = false;
+flush:
+				// finish up the line
+				buf[pos - shift] = '\0';
```
trufae, line 264: is shift always smalle than pos?

src/text.c
```diff
@@ -0,0 +1,358 @@
...
+	FILE *f = fopen (file, "w");
+	if (!f) {
+		return false;
+	}
+	bool r = sdb_text_fsave (s, f, sort);
+	fclose (f);
+	return r;
+}
+
+SDB_API bool sdb_text_fload(Sdb *s, FILE *f) {
```
trufae, line 200: this function is too long. can be split into small functions to make the main parsing loop easier to read

test/unit/test_sdb.c
```diff
@@ -207,6 +207,146 @@ bool test_sdb_copy() {
...
+	char buf[2048];
+	if (sdb_diff_format (buf, sizeof (buf), diff) < 0) {
+		return;
+	}
+	printf ("%s\n", buf);
+}
+
+bool test_sdb_text_load_simple() {
+#ifdef __linux__
+	FILE *f = fmemopen ((void *)text_ref_simple, strlen (text_ref_simple), "r");
```
trufae, line 318: mmap+fdopen maybe?

* https://github.com/Arryboom/fmemopen_windows
* https://github.com/libconfuse/libconfuse/blob/master/src/fmemopen.c

i understand its just for a test. but i do think parsing from memory is quite common usecase that we may provide apis to make sdb_text_load() work with just the pointers

trufae (changes requested):

src/ls.c
thestr4ng3r: It's the same commit, will disappear on rebase.

src/text.c
thestr4ng3r: No, I always need to print a slash, even when path is empty. So I have to do it outside the loop and skip it in the first loop iteration, which is what the `first` is needed for.

src/text.c
thestr4ng3r: Yes, it doesn't, it would handle them as regular data characters. While I think non-unix newlines should die and rot in hell, it's not too hard to ignore them and make `\r` for escaping them in data.

src/text.c
thestr4ng3r: > mmm is it ever possible to have more than one name in path being passed to text_fsave?

Yes, because `text_fsave` recurses. When you have the path "/some/path/here", the list will be `["some", "path", "here"]`. This is to avoid clashing with '/' chars in path tokens.

ret2libc: I see now yeah :+1:

src/text.c
thestr4ng3r: It is always <= pos, yes.

src/text.c
thestr4ng3r: switch on the state would need code duplication, but I think after splitting it into multiple functions, it is better anyway.

test/unit/test_sdb.c
thestr4ng3r: Yeah, having it work on memory can be useful indeed. Still, I would not want to lose the ability to use this streaming parser from a file without having to slurp the entire file so something like these alternative fmemopen impls might be interesting in the future.

thestr4ng3r:
> I'm not super familiar with SDB internals, but by reading the issues it said SDB shall be able to automatically detect if loading a text format or a "compiled" one. Where is this thing done?

Currently nowhere. The regular cdb access is embedded into the Sdb structure itself somehow using mmap instead of loading while this is just a parser that loads a file into memory so it's a somewhat different approach. But a magic detection could be built on top.

thestr4ng3r:
ping @trufae

src/text.c
```diff
@@ -0,0 +1,410 @@
...
+
+	// k=v entries
+	SdbList *l = sdb_foreach_list (s, sort);
+	SdbKv *kv;
+	SdbListIter *it;
+	ls_foreach (l, it, kv) {
+		const char *k = sdbkv_key (kv);
+		const char *v = sdbkv_value (kv);
+		// escape leading '/'
+		if (*k == '/') {
```
ret2libc, line 159: move this into `write_key`.

src/text.c
```diff
@@ -0,0 +1,410 @@
...
+		ctx->bufsz = newsz;
+		char *newbuf = realloc (ctx->buf, ctx->bufsz);
+		if (!newbuf) {
+			return false;
+		}
+		ctx->buf = newbuf;
+	}
+	// read full buffer (leave 1 for null-terminating)
+	size_t red = fread (ctx->buf + ctx->pos, 1, ctx->bufsz - 1 - ctx->pos, ctx->f);
+	if (!red && !feof (ctx->f)) {
```
ret2libc, line 249: I'm not sure `feof` requires `red == 0`. At least the man page doesn't say so, so I guess there can be cases where it reads few bytes and set the status of the file to EOF.

src/text.c
```diff
@@ -0,0 +1,410 @@
...
+			ret = false;
+			break;
+		}
+		// process buffer
+		if (ctx.eof) {
+			load_flush_line (&ctx);
+			break;
+		}
+		while (ctx.pos < ctx.buf_filled) {
+			if (ctx.buf[ctx.pos] == '\n') {
```
ret2libc, line 389: this is just another character to process, so this `if` imo should go in `load_process_single_char`.

thestr4ng3r:
ping

src/text.c
```diff
@@ -0,0 +1,410 @@
...
+// to be called at the end of a line.
+// save all the data processed from the line into the database.
+static void load_flush_line(LoadCtx *ctx) {
+	ctx->unescape = false;
+	// finish up the line
+	ctx->buf[ctx->pos - ctx->shift] = '\0';
+	switch (ctx->state) {
+	case STATE_PATH: {
+		ls_push (ctx->path, (void *)ctx->token_begin);
+		// TODO: remove ctx->token_begin = ctx->pos + 1;
```
ret2libc, line 268: what does this mean?

test/unit/test_sdb.c
```diff
@@ -219,6 +374,10 @@ int all_tests() {
 	mu_run_test (test_sdb_list_big);
 	mu_run_test (test_sdb_foreach_filter);
 	mu_run_test (test_sdb_copy);
+	mu_run_test (test_sdb_text_save_simple);
+	mu_run_test (test_sdb_text_load_simple);
+	mu_run_test (test_sdb_text_save);
+	mu_run_test (test_sdb_text_load);
```
ret2libc, line 380: could you also try running some tests with invalid projects to load? As this will be used for projects, I think we should be careful to ensure the loading of a SDB text is as safe as possible.

ret2libc (commented):
It looks good to me, though there are still some comments and I would also love to see how it fails in case of malformed input.

src/sdb.h
```diff
@@ -206,6 +206,12 @@ SDB_API bool sdb_disk_insert(Sdb* s, const char *key, const char *val);
 SDB_API bool sdb_disk_finish(Sdb* s);
 SDB_API bool sdb_disk_unlink(Sdb* s);
 
+/* plaintext sdb files */
+SDB_API bool sdb_text_fsave(Sdb *s, FILE *f, bool sort);
```
trufae, line 210: FILE is not used in SDB until now, if we have the same callbacks for bin and text we can easily wrap them into a generic struct with callbacks to select different backends

src/text.c
```diff
@@ -0,0 +1,410 @@
...
+	char *buf;
+	size_t buf_filled; // how many bytes of buf are currently filled with data from the file
+	Sdb *root_db;
+	Sdb *cur_db; // current namespace, changes when encountering a path line
+	// can't use pointers here because buf may be moved on realloc
+	size_t pos; // current processing position in the buffer
+	size_t token_begin; // beginning of the currently processed token in the buffer
+	size_t shift; // amount to shift chars to the left (from unescaping)
+	SdbList/*<size_t>*/ *path;
+	enum { STATE_NEWLINE, STATE_PATH, STATE_KEY, STATE_VALUE } state;
```
trufae, line 226: uhm better to move this enum definition outside the struct

src/text.c
```diff
@@ -0,0 +1,410 @@
...
+	// realloc if no space
+	if (ctx->pos >= ctx->bufsz - 1) {
+		size_t newsz = ctx->bufsz + ctx->bufsz;
+		if (newsz < ctx->bufsz) {
+			return false;
+		}
+		ctx->bufsz = newsz;
+		char *newbuf = realloc (ctx->buf, ctx->bufsz);
+		if (!newbuf) {
+			return false;
```
trufae, line 243: ctx->bufsz is inconsistent with ctx->buf at this moment. move the assignemnt below

trufae (changes requested):

trufae:
Some questions/thoughs:

- saving an sdb as plaintext is just sdb foo.sdb > foo.txt
- loading sdb foo.txt shouldnt work with your PR, afaik. that should be automatically detected
- public APIs shouldnt involve FILE*

src/text.c
thestr4ng3r: Yes, that is correct. `feof()` will most of the time be set after reading some bytes. But then in the next iteration, `red` will always be `0` and `feof()` still true so it will exit there. This just makes the code a bit simpler.

src/text.c
thestr4ng3r: Nothing, it's a leftover

thestr4ng3r:
>     * public APIs shouldnt involve FILE*

Should I make it int fds instead?

thestr4ng3r:
>     * saving an sdb as plaintext is just sdb foo.sdb > foo.txt

Currently, not exactly. What this does right now is this `sdb_grep_dump()` function in `main.c` which can do different formats, also greps and outputs without escaping and without namespaces. So re-reading that will not always work.

If we wanted to use this "real" plaintext sdb format, we would have to first build the result in memory, because of the grepping, then dump it out and add some more code for json. Should we do that?

>     * loading sdb foo.txt shouldnt work with your PR, afaik. that should be automatically detected

How do we want to do that? CDB afaik has no magic. We could make the plaintext format however enforce that it must always start with a single `/` line but that might get unreliable when a user not knowing this adds whitespace before or, even worse, you have a cdb that happens to start with `/`.

test/unit/test_sdb.c
thestr4ng3r: There actually aren't too many cases of invalid sdb text. There are some things that are nonsense, but the parser just handles like `\r` newlines, multiple `/` or escaping chars that don't have to be escape in the respective scope.
The only real "invalid" thing is a line that does not start with `/` and does not contain a non-escaped `=` which would be just ignored.
I've added some tests for these things.

ret2libc: What about things like `\` and then a newline? I think the newline would become part of the token, right? Is that ok?

thestr4ng3r: Good idea. I think it will ignore the `\` and treat the newline as record separator, but I have to check.

thestr4ng3r: Yep, added a test and it just ignores it.

src/text.c
trufae: i can confirm this behaviour despite not being documented

thestr4ng3r: It is documented, just phrased differently and in different places.

trufae:
What i was thinking as an alternative is to read the first 32 or 64 bytds from the file and follow some checks:

* plaintext if everything is printable and contains at least one = and one newline
* its binary if there are several aligned null bytes

trufae:
The problem witth adding a header in cdb is that it breaks the whole fun of just being a raw thing you map in memory and alignments matter. otherwise adding a +4 to skip the header will involve many ugly changes. i was thinking in adding a footer instead. but there was no use for it... yet

trufae:
ill submit a proposal api based on this idea of isSdb() {return isBinarySdb()||isTextSdb()}

thestr4ng3r:
> What i was thinking as an alternative is to read the first 32 or 64 bytds from the file and follow some checks:
> 
>     * plaintext if everything is printable and contains at least one = and one newline
> 
>     * its binary if there are several aligned null bytes

That would be quite fuzzy and unreliable so more of a thing for just the user interface maybe. But usually it will probably succeed. For checking plaintext, I would also check for `/`.

So what do we do about the `FILE *`s? Just replace by fds?

thestr4ng3r:
so wat do

thestr4ng3r:
Now using fds.

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+		switch (ctx->buf[ctx->pos]) {
+			case 'n':
+				raw_char = '\n';
+				break;
+			case 'r':
+				raw_char = '\r';
+				break;
+			default:
+				raw_char = ctx->buf[ctx->pos];
+			break;
```
ret2libc, line 363: wrong indentation

ret2libc (commented):
you commit `.tmp` file

ret2libc (approved):
LGTM, awesome code

thestr4ng3r:
> you commit `.tmp` file

shit. Thanks for noting.

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+			first = false;
+		} else {
+			write (fd, "/", 1);
+		}
+		// write and escape the path
+		const char *p = path_token;
+		const char *n = p;
+		while (*n) {
+			switch (*n) {
+			case '\\':
```
trufae, line 87: this char replacing thing is done in 3 places, and its duplicating the same code for each case, when it can be simplified to just replace one char after the backslash and implemented in a single helper function. can you do this refactoring?

* write_path
* write_key
* write_value

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+			write (fd, "\\r", 2);
+			break;
+		}
+		n++;
+	}
+	FLUSH;
+}
+
+static bool text_save(Sdb *s, int fd, bool sort, SdbList *path) {
+	//path
```
trufae, line 166: space after the two slashes

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+		.root_db = s,
+		.cur_db = s,
+		.pos = 0,
+		.token_begin = 0,
+		.shift = 0,
+		.path = ls_new (),
+		.state = STATE_NEWLINE,
+		.unescape = false
+	};
+	if (!ctx.buf || !ctx.path) {
```
trufae, line 407: if (ctx. buf && ctx.path) { ...} so you avoid having dupplicated free/ls_free. btw thise handmade free and struct initialization can be done in a separate functions, so it makes it cleaner, mainly because ctx just needs the fd to be passed, all the other stuff is 0 or allocated

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+		free (ctx.buf);
+		ls_free (ctx.path);
+		return false;
+	}
+	bool ret = true;
+	while (true) {
```
trufae, line 413: i prefer for(;;) for endless loops. because while(true) implies having a condition to check (which compilers may obvious optimize accordingly), but meh

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+		}
+		char *newbuf = realloc (ctx->buf, newsz);
+		if (!newbuf) {
+			return false;
+		}
+		ctx->buf = newbuf;
+		ctx->bufsz = newsz;
+	}
+	// read full buffer (leave 1 for null-terminating)
+	int red = read (ctx->fd, ctx->buf + ctx->pos, ctx->bufsz - 1 - ctx->pos);
```
trufae, line 271: using slurp() you dont need to handle all the buffer caching by yourself. so i would vote for simpler implementation by just reading the whole file in memory (or plain mmap), and make the text_load() function take the filename as argument. itt will be faster and simpler

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+			ctx->pos++;
+			return;
+		} else {
+			ctx->state = STATE_KEY;
+		}
+	}
+
+	if (ctx->unescape) {
+		char raw_char;
+		switch (ctx->buf[ctx->pos]) {
```
trufae, line 354: cache ctx->buf[ctx->pos] in char curchar, this is being checked several times below and you dont need to dereference 2 pointers on every if

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+		case 'n':
```
trufae, line 355: imho this is cleaner (and faster, no need for a table lookup to chk 2 chars) with a single line with ternary operator:

`char raw_char = curchar == 'n'? '\n': (curchar == 'r')? '\r': curchar`

Also, what happens with all the other escapable chars. considering only newlines to be what should be escaped is not very flexible. imho the escaping function should be ripped from r2 and support \t and \x at least

test/unit/test_sdb.c
```diff
@@ -207,6 +207,235 @@ bool test_sdb_copy() {
...
+	return fd;
+}
+
+bool test_sdb_text_save_simple() {
+	Sdb *db = text_ref_simple_db ();
+
+	int fd = tmpfile_new (".text_save_simple", NULL, 0);
+	bool succ = sdb_text_save_fd (db, fd, true);
+	lseek (fd, 0, SEEK_SET);
+	char buf[0x1000];
```
trufae, line 333: use a define at the begining of the file instead of 0x1000

test/unit/test_sdb.c
```diff
@@ -207,6 +207,235 @@ bool test_sdb_copy() {
...
+	sdb_free (ref_db);
+	sdb_free (db);
+	mu_assert_true (eq, "load correct");
+	mu_end;
+}
+
+bool test_sdb_text_load_broken() {
+	int fd = tmpfile_new (".text_load_broken", text_ref_broken, strlen (text_ref_broken));
+	Sdb *db = sdb_new0 ();
+	bool succ = sdb_text_load_fd (db, fd);
```
trufae, line 426: imho sdb_load() should work for plain and bin databases. exposing text/binary APIs is not good. i would even prefer adding a parameter to the sdb_load api to specify the file format (TEXT, BINARY, GUESS). What do you think?

trufae (changes requested):

src/text.c
ret2libc: :-1: this is not cleaner at all to me. You just write everything in one single line, it doesn't make it any cleaner nor faster (i bet *ANY* compiler can compile the switch case in the best way, without requiring a table lookup).

trufae: then just wrap this into a function

src/text.c
```diff
@@ -0,0 +1,441 @@
...
+	if (!path) {
+		return false;
+	}
+	bool r = text_save (s, fd, sort, path);
+	ls_free (path);
+	return r;
+}
+
+SDB_API bool sdb_text_save(Sdb *s, const char *file, bool sort) {
+	int fd = open (file, O_WRONLY | O_CREAT | O_TRUNC);
```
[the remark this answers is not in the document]
thestr4ng3r: Yeah it treats the inferior kind of newline as regular newlines too so it doesn't matter anymore.

src/text.c
thestr4ng3r: I would like to avoid slurp since these text files can become quite large. But I will try to do it with mmap+cow, which might even work very similar to cdb and then I can embed it in the Sdb struct if possible.

test/unit/test_sdb.c
thestr4ng3r: You mean sdb_open?

src/text.c
```diff
@@ -163,7 +163,7 @@ static void write_value(int fd, const char *v) {
 }
 
 static bool text_save(Sdb *s, int fd, bool sort, SdbList *path) {
-	//path
+	// path
```
trufae, line 166: this comment can be removed

thestr4ng3r: Every comment can be removed, but I would like to keep this one since it splits the function in three distinct sections.

src/text.c
thestr4ng3r: A function would be quite ugly here since you need to pass many pointers to local vars that have to be changed. I can do some more macro magic though to make it look nicer. Whether that is better then is a matter of taste though.

thestr4ng3r: magic applied now, please check again

src/main.c
```diff
@@ -256,9 +256,8 @@ static int insertkeys(Sdb *s, const char **args, int nargs, int mode) {
 }
 
 static int createdb(const char *f, const char **args, int nargs) {
-	char *line, *eq;
 	s = sdb_new (NULL, f, 0);
-	if (!s || !sdb_disk_create (s)) {
+	if (!s) {
```
trufae, line 260: uhm, createdb is no longer creating the database? dont we have tests for this?

src/sdb.h
```diff
@@ -206,6 +206,12 @@ SDB_API bool sdb_disk_insert(Sdb* s, const char *key, const char *val);
 SDB_API bool sdb_disk_finish(Sdb* s);
 SDB_API bool sdb_disk_unlink(Sdb* s);
 
+/* plaintext sdb files */
+SDB_API bool sdb_text_save_fd(Sdb *s, int fd, bool sort);
```
trufae, line 210: there's still an fd here

src/text.c
```diff
@@ -0,0 +1,394 @@
...
+	SdbListIter *it;
+	const char *path_token;
+	bool first = true;
+	ls_foreach (path, it, path_token) {
+		if (first) {
+			first = false;
+		} else {
+			write (fd, "/", 1);
+		}
+		ESCAPE_LOOP (fd, path_token,
```
trufae, line 100: this is not solving anything of what i commented, just hiding all the complexity behind macros

src/text.c
```diff
@@ -0,0 +1,394 @@
...
+		text_save (ns->sdb, fd, sort, path);
+		ls_pop (path);
+	}
+	if (l != s->ns) {
+		ls_free (l);
+	}
+
+	return true;
+}
+#undef FLUSH
```
trufae, line 170: this is being unused one function before

src/text.c
```diff
@@ -0,0 +1,394 @@
...
+	}
+	default:
+		break;
+	}
+	// prepare for next line
+	ctx->shift = 0;
+	ctx->state = STATE_NEWLINE;
+}
+
+static inline char unescape_raw_char (char c) {
```
trufae, line 259: also unescape tabs. also add the reverse function and use it instead of the macros above

src/main.c
trufae: maybe the behaviour is the same, and i didnt chkd but i dont think sdb_new is creating the file if it doesnt exist

trufae (changes requested):

src/main.c
thestr4ng3r: It will be created in `sdb_sync()` later, so the behavior is the same.

src/sdb.h
thestr4ng3r: That is correct, because the saving directly streams out the data without having to accumulate it in memory first. How would you solve it instead?

src/text.c
thestr4ng3r: That's what I wrote.
> A function would be quite ugly here since you need to pass many pointers to local vars that have to be changed. I can do some more macro magic though to make it look nicer. Whether that is better then is a matter of taste though.

src/text.c
thestr4ng3r: Nope, there is no direct inverse.

src/main.c
trufae: Not really because the early creation may be used for locking purposes. But as long as rhere are no tests for this its hard to say the current behaviour is good

src/text.c
```diff
@@ -0,0 +1,396 @@
...
+}
+
+static void write_key(int fd, const char *k) {
+	// escape leading '/'
+	if (*k == '/') {
+		write (fd, "\\", 1);
+	}
+	ESCAPE_LOOP (fd, k,
+		ESCAPE ('\\', "\\", 1);
+		ESCAPE ('=', "=", 1);
```
trufae, line 116: Escape tab? (In all 3 functions) escaping things

thestr4ng3r: I strongly prefer not to because it's unnecessary and just uses more memory and slows down parsing.

thestr4ng3r: ping @trufae

trufae: i dont understand how replacing \t with \\t will be a memory problem at all. but anyway, tab is a valid char for sdb, so lets keep it like this

src/text.c
trufae: the function signature shouldnt hgave a space before (

trufae: there's a inverse, but takes longer to explain than to implement, so not going to discuss further here. its easier for me to explain in another PR after merging this

trufae: also this char c can be const char c, because it wont be modified inside. but thats just little picky details

src/sdb.h
trufae: but this should be only an internal function, not exposed publickly. windows shouldnt be using this integer file descriptors things. and passing the filename should be enough imho.

trufae:
Sorry i had my review without submitting it

src/text.c
```diff
@@ -0,0 +1,396 @@
...
+#undef ESCAPE_LOOP
+#undef ESCAPE
+
+static bool text_save(Sdb *s, int fd, bool sort, SdbList *path) {
+	// path
+	write_path (fd, path);
+	write (fd, "\n", 1);
+
+	// k=v entries
+	SdbList *l = sdb_foreach_list (s, sort);
```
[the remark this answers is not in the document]
thestr4ng3r: It's done to sort the keys. I don't like the fact that you need an extra list either, but for sorting it's necessary. I've added another path using callback-iterating and added a test for it.

trufae (approved):

test/unit/test_sdb.c
```diff
@@ -346,6 +360,26 @@ bool test_sdb_text_save_simple() {
 	mu_end;
 }
 
+bool test_sdb_text_save_simple_unsorted() {
+	Sdb *db = text_ref_simple_db ();
+
+	int fd = tmpfile_new (".text_save_simple_unsorted", NULL, 0);
+	bool succ = sdb_text_save_fd (db, fd, false);
+	lseek (fd, 0, SEEK_SET);
+	char buf[TEST_BUF_SZ];
```
trufae, line 369: you can use = {0} and remove the memset below

PR #222, opened 2020-10-12
Remove some useless leaking code

thestr4ng3r:


test/test.txt
```diff
@@ -0,0 +1,3 @@
+fuck=you
```
trufae, line 1: please reword the test file to not contain words that may offend people against anal sex

trufae (changes requested):

PR #224, opened 2020-10-23
Remove unused code

XVilka:
**Detailed description**

There is no reason to keep it in the repository - if anything it could be extracted from the history.

**Test plan**

All green


ret2libc (approved):

trufae:
Sdb uses to be a datastructure collection library and tool, all those algorithms were moved to Attic after deciding other implementations were more interesting for the actual sdb use case. despite Attic comes from the CVS era, i would like to have them somewhere else so its easy for anyone to find and use them if they are needed for their project, i spent quite a lot of time cleaning up and testing all of those to be easy to use and clean APIs.

I'll just create a repo with the contents of this Attic in my repository and make the dist tarballs 25KB smaller!

trufae:
Done

PR #225, opened 2020-10-28
Add support for `local` meson option

ret2libc:
**Detailed description**

Similar to radare2, it allows to install sdb in a local path and set rpath so that you don't have to mess with LD_LIBRARY_PATH & co.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -64,12 +64,19 @@ libsdb_sources = [
 ]
 
 sdb_inc = include_directories(['.', 'src'])
+rpath_lib = ''
+rpath_exe = ''
+if get_option('local') and get_option('default_library') == 'shared'
+  rpath_lib = '$ORIGIN'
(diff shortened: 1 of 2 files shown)


thestr4ng3r (approved):

meson.build
```diff
@@ -64,12 +64,19 @@ libsdb_sources = [
 ]
 
 sdb_inc = include_directories(['.', 'src'])
+rpath_lib = ''
+rpath_exe = ''
+if get_option('local') and get_option('default_library') == 'shared'
+  rpath_lib = '$ORIGIN'
+  rpath_exe = '$ORIGIN/../' + get_option('libdir')
```
kazarmy, line 71: I actually prefer that you copy r2's `meson.build` exactly but nbd.

kazarmy (approved):

PR #227, opened 2020-10-28
Fix heap buffer overflow

pelijah:
**Detailed description**

Reserve extra byte for terminating null to prevent heap buffer overflow (https://github.com/radareorg/sdb/blob/master/src/text.c#L229) in last call of `load_flush_line()` (https://github.com/radareorg/sdb/blob/master/src/text.c#L364)

**Test plan**

...

**Closing issues**

...

### DIFF
--- a/src/text.c
+++ b/src/text.c
@@ -382,7 +382,7 @@ SDB_API bool sdb_text_load(Sdb *s, const char *file) {
 		goto beach;
 	}
 #else
-	char *x = calloc (1, st.st_size);
+	char *x = calloc (1, st.st_size + 1);
 	if (!x) {
 		goto beach;
 	}


pelijah:
@thestr4ng3r

thestr4ng3r:
Same issue as in #226. While your change fixes the issue for loading files without mmap, it is still technically wrong with how the API of `sdb_text_load_buf()` is meant to be used (the buffer is allowed to be only of the given size) and will also fail when loading a file which has a size of exactly a multiple of the pagesize with mmap.

pelijah:
Meh...  I’m not even the first...

thestr4ng3r:
Funny that we both found the bug within one hour after it has existed for 16 days.

PR #231, opened 2020-12-15
Compile a native sdb binary when cross building

ret2libc:
**Detailed description**

sdb binary may be used by the build machine to compile .sdb files. In
that case it is necessary to have the sdb binary compiled for the build
machine and not for the host one, as it is by default.

**Test plan**

Create a file `cross-file.txt` with:
```
[properties]
exe_wrapper = 'wine'
```
and then run `meson build --cross-file cross-file.txt ; ninja -C build`. It should work and it should create another binary called sdb_native.

**Closing issues**

None

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -118,6 +118,17 @@ sdb_exe = executable('sdb', 'src/main.c',
   implicit_include_directories: false
 )
 
+if meson.is_cross_build()
+  sdb_native_exe = executable('sdb_native', 'src/main.c',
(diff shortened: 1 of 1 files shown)


trufae:
Can you update the README with this information? thanks!

trufae:
i'll document that

trufae:
Done in https://github.com/radareorg/sdb/commit/943b7c63dc26d0747fe5f04a0a1b926480f9e0e7

PR #235, opened 2020-12-28
 SDB: In Soviet Russia, C has booleans ##refactor

ratijas:
**Detailed description**

Change arguments type from int to bool. Those are internal static function, so no other code could possibly be affected. There are booleans all over the place in that file, so we are not losing any portability.

PS Editing raw `*.patch`es to match new paths is a surprisingly valid technique.

**Test plan**

None

**Closing issues**

Originally (mistakenly) posted at: https://github.com/radareorg/radare2/pull/18120

### DIFF
--- a/src/sdb.c
+++ b/src/sdb.c
@@ -546,7 +546,7 @@ SDB_API void sdbkv_free(SdbKv *kv) {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


trufae:
thanks!

PR #242, opened 2021-09-19
Rework sdb_itoa buffer management to silence GCC warning

hno:
make a clearer separation between temporary and allocated string buffer
eleminating gcc warning about function may return address of local variable

**Checklist**

- [x] Closing issues: radare2 #18428
- [x] Mark this if you consider it ready to merge
- [ ] I've added tests (optional)
- [ ] I wrote some documentation

**Description**

See above. It's a cosmetic bugfix. Nothing new to document or add.


trufae:
Also updated in r2! thanks!

PR #244, opened 2022-01-17
Remove TLS and make sdb thread-safe

meme:
**Checklist**

- [ ] Closing issues: #issue
- [ ] Mark this if you consider it ready to merge
- [ ] I've added tests (optional)
- [ ] I wrote some documentation

**Description**

This makes sdb thread-safe without the need for TLS.

* Nuke sdb_fmt
* Mark goto tables as const
* Mark lookup table as const
* Move nextcas into SdbKv
* Remove global hook (cannot be used without TLS, but we want to avoid
  that)
* Fix lock path to not use a static buffer

### DIFF
--- a/src/journal.c
+++ b/src/journal.c
@@ -6,29 +6,39 @@
 #include <unistd.h>
 #endif
 
-static const char *sdb_journal_filename(Sdb *s) {
-	return (s && s->name)
-		? sdb_fmt ("%s.journal", s->name)
-		: NULL;
+static bool sdb_journal_filename(Sdb *s, char *path, size_t path_size) {
+	if (!s || !s->name) {
+		return false;
+	}
+
+	int res = snprintf (path, path_size, "%s.journal", s->name);
+	if (res < 0 || (size_t)res >= path_size) {
+		return false;
+	}
+
+	return true;
 }
 
 SDB_API bool sdb_journal_close(Sdb *s) {
+	char filename[SDB_MAX_PATH];
 	if (s->journal == -1) {
 		return false;
 	}
 	close (s->journal);
 	s->journal = -1;
-	unlink (sdb_journal_filename (s));
+	if (!sdb_journal_filename (s, filename, sizeof (filename))) {
+		return false;
+	}
+	unlink (filename);
 	return true;
 }
 
 SDB_API bool sdb_journal_open(Sdb *s) {
-	const char *filename;
+	char filename[SDB_MAX_PATH];
 	if (!s || !s->name) {
 		return false;
 	}
-	filename = sdb_journal_filename (s);
-	if (!filename) {
+	if (!sdb_journal_filename (s, filename, sizeof (filename))) {
 		return false;
 	}
 	close (s->journal);
@@ -81,10 +91,13 @@ SDB_API int sdb_journal_load(Sdb *s) {
 }
 
 SDB_API bool sdb_journal_log(Sdb *s, const char *key, const char *val) {
+	char str[SDB_MAX_PATH];
 	if (s->journal == -1) {
 		return false;
 	}
-	const char *str = sdb_fmt ("%s=%s\n", key, val);
+	if (snprintf (str, sizeof (str), "%s=%s\n", key, val) < 0) {
+		return false;
+	}
 	int len = strlen (str);
 	if (write (s->journal, str, len) != len) {
 		return false;
@@ -103,10 +116,10 @@ SDB_API bool sdb_journal_clear(Sdb *s) {
 }
 
 SDB_API bool sdb_journal_unlink(Sdb *s) {
-	const char *filename = sdb_journal_filename (s);
-	sdb_journal_close (s);
-	if (filename) {
-		return !unlink (filename);
+	char filename[SDB_MAX_PATH];
+	if (!sdb_journal_filename (s, filename, sizeof (filename))) {
+		return false;
 	}
-	return false;
+	sdb_journal_close (s);
+	return !unlink (filename);
 }
--- a/src/query.c
+++ b/src/query.c
@@ -207,12 +207,14 @@ repeat:
 	quot = NULL;
 	json = NULL;
 	if (*p == '#') {
+		char buffer[16];
 		p++;
 		next = strchr (p, ';');
 		if (next) {
 			*next = 0;
 		}
-		out_concat (sdb_fmt ("0x%08x\n", sdb_hash (p)));
+		(void)snprintf (buffer, sizeof (buffer), "0x%08x\n", sdb_hash (p));
+		strbuf_append (out, buffer, 1);
 		if (next) {
 			*next = ';';
 		}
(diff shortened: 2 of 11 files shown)


src/journal.c
```diff
@@ -103,10 +115,10 @@ SDB_API bool sdb_journal_clear(Sdb *s) {
 }
 
 SDB_API bool sdb_journal_unlink(Sdb *s) {
-	const char *filename = sdb_journal_filename (s);
-	sdb_journal_close (s);
-	if (filename) {
-		return !unlink (filename);
+	char filename[256];
```
trufae, line 118:
```suggestion
	char filename[SDB_MAX_PATH];
```

which should be the same as MAX_PATH from limits.h, or 256 if not defined (windows)

src/journal.c
```diff
@@ -6,29 +6,38 @@
...
-static const char *sdb_journal_filename(Sdb *s) {
-	return (s && s->name)
-		? sdb_fmt ("%s.journal", s->name)
-		: NULL;
+static int sdb_journal_filename(Sdb *s, char *path, size_t path_size) {
+	if (!s || !s->name) {
+		return -1;
+	}
+
+	if (snprintf (path, path_size, "%s.journal", s->name) < 0) {
```
trufae, line 14:
```suggestion
	int res = snprintf (path, path_size, "%s.journal", s->name);
	if (res < 0 || res >= path_size) {
```

src/journal.c
```diff
@@ -6,29 +6,38 @@
 #include <unistd.h>
 #endif
 
-static const char *sdb_journal_filename(Sdb *s) {
-	return (s && s->name)
-		? sdb_fmt ("%s.journal", s->name)
-		: NULL;
+static int sdb_journal_filename(Sdb *s, char *path, size_t path_size) {
```
trufae, line 9:
```suggestion
static bool sdb_journal_filename(Sdb *s, char *path, size_t path_size) {
```

src/journal.c
```diff
@@ -6,29 +6,38 @@
...
+	if (!s || !s->name) {
+		return -1;
```
trufae, line 11:
```suggestion
		return false;
```

src/journal.c
```diff
@@ -6,29 +6,38 @@
...
+	}
+
+	if (snprintf (path, path_size, "%s.journal", s->name) < 0) {
+		return -1;
```
trufae, line 15:
```suggestion
		return false;
```

src/journal.c
```diff
@@ -6,29 +6,38 @@
...
+	}
+
+	return 0;
```
trufae, line 18:
```suggestion
	return true;
```

src/journal.c
```diff
@@ -6,29 +6,38 @@
...
 
 SDB_API bool sdb_journal_close(Sdb *s) {
+	char filename[256];
 	if (s->journal == -1) {
 		return false;
 	}
 	close (s->journal);
 	s->journal = -1;
-	unlink (sdb_journal_filename (s));
+	if (sdb_journal_filename (s, filename, sizeof (filename)) == -1) {
```
trufae, line 28:
```suggestion
	if (!sdb_journal_filename (s, filename, sizeof (filename))) {
```

src/journal.c
```diff
@@ -6,29 +6,38 @@
...
 
 SDB_API bool sdb_journal_open(Sdb *s) {
-	const char *filename;
+	char filename[256];
 	if (!s || !s->name) {
 		return false;
 	}
-	filename = sdb_journal_filename (s);
-	if (!filename) {
+	if (sdb_journal_filename (s, filename, sizeof (filename)) == -1) {
```
trufae, line 40:
```suggestion
	if (!sdb_journal_filename (s, filename, sizeof (filename))) {
```

src/query.c
```diff
@@ -207,12 +207,14 @@ SDB_API char *sdb_querys(Sdb *r, char *buf, size_t len, const char *_cmd) {
...
 	json = NULL;
 	if (*p == '#') {
+		char buffer[256];
 		p++;
 		next = strchr (p, ';');
 		if (next) {
 			*next = 0;
 		}
-		out_concat (sdb_fmt ("0x%08x\n", sdb_hash (p)));
+		snprintf (buffer, sizeof(buffer), "0x%08x\n", sdb_hash (p));
```
trufae, line 216:
```suggestion
		(void)snprintf (buffer, sizeof (buffer), "0x%08x\n", sdb_hash (p));
```

src/query.c
```diff
@@ -207,12 +207,14 @@ SDB_API char *sdb_querys(Sdb *r, char *buf, size_t len, const char *_cmd) {
 	quot = NULL;
 	json = NULL;
 	if (*p == '#') {
+		char buffer[256];
```
trufae, line 210: the snprintf below can't fill 256 bytes, allocating 16 should be enough
```suggestion
		char buffer[16];
```

src/sdb.c
```diff
@@ -134,13 +129,16 @@ SDB_API Sdb* sdb_new(const char *path, const char *name, int lock) {
 
 // XXX: this is wrong. stuff not stored in memory is lost
 SDB_API void sdb_file(Sdb* s, const char *dir) {
+	char buf[128];
```
trufae, line 132:
```suggestion
	char buf[SDB_MAX_PATH];
```

test/unit/test_sdb.c
```diff
@@ -89,14 +85,17 @@ bool test_sdb_delete_none(void) {
 
 bool test_sdb_delete_alot(void) {
 	Sdb *db = sdb_new (NULL, NULL, false);
+	char buffer[256];
```
trufae, line 88:
```suggestion
	char buffer[32];
```

"key.%d" can be 4 + 11 + 1 as maximum, so 16 should be fine, but maybe better to give few more just to be safe in case of future changes

test/unit/test_sdb.c
```diff
@@ -89,14 +85,17 @@ bool test_sdb_delete_none(void) {
...
 	const int count = 2048;
 	int i;
 
 	for (i = 0; i < count; i++) {
-		sdb_set (db, sdb_fmt ("key.%d", i), "bar", 0);
+		snprintf (buffer, sizeof (buffer), "key.%d", i);
```
trufae, line 93:
```suggestion
		(void)snprintf (buffer, sizeof (buffer), "key.%d", i);
```

test/unit/test_sdb.c
```diff
@@ -89,14 +85,17 @@ bool test_sdb_delete_none(void) {
...
+		sdb_set (db, buffer, "bar", 0);
 	}
 	for (i = 0; i < count; i++) {
-		sdb_unset (db, sdb_fmt ("key.%d", i), 0);
+		snprintf (buffer, sizeof (buffer), "key.%d", i);
```
trufae, line 97:
```suggestion
		(void)snprintf (buffer, sizeof (buffer), "key.%d", i);
```

src/journal.c
```diff
@@ -6,29 +6,38 @@
...
+		return false;
+	}
+	unlink (filename);
 	return true;
 }
 
 SDB_API bool sdb_journal_open(Sdb *s) {
-	const char *filename;
+	char filename[256];
```
trufae, line 36:
```suggestion
	char filename[SDB_MAX_PATH];
```

trufae (changes requested):

